Issue #12, opened 2017-08-08
Failing property in PropCheck.Test.Movies


evnu:
With 35bfdaf and Elixir v1.5, the "server works fine" property in `test/movie_test.exs` fails here:

```
  1) property server works fine (PropCheck.Test.Movies)
     test/movie_test.exs:19
     Property server works fine failed. Counter-Example is:
     [[{:set, {:var, 1}, {:call, PropCheck.Test.MovieServer, :ask_for_popcorn, []}}]]
     
     code: nil
     stacktrace:
       test/movie_test.exs:19: (test)
```

alfert:
This has apparently nothing to do with your changes but more with Elixir 1.5 - for whatever reason. I will take a  look at that.

alfert:
Hmm, fails also in Elixir 1.4.5 under OTP 18 on travis.

alfert:
Fixed in #15. I have no clue how this test has ever run, since starting the movie server GenServer was simply wrong.

PR #14, opened 2017-08-12
Test Elixir v1.3, v1.4 and v1.5 on travis-ci

evnu:
This commit changes `.travis.yml` in order to check multiple Elixir versions, in case that `propcheck` should support more than only the newest release.

Note that the commit removes `notifications`, as I do not know if a failing build on my branch would then send unneeded notifications. If this pull request will be accepted, I will re-add the notification again first.

Also note that I stripped out `otp_release`, as Travis will take care of picking a suitable release. Further, the commit does not specify the PATCH release version of Elixir, allowing Travis to simply use the newest release with the provided MAJOR.MINOR version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,8 @@
 language: elixir
-notifications:
-  recipients:
-    - klaus.alfert@googlemail.com
-otp_release:
-  - 18.2.1
 elixir:
-  - 1.3.0
+  - 1.3
+  - 1.4
+  - 1.5
 before_script:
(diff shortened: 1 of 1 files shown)


evnu:
Note that the tests fail for v1.4 and v1.5 due to #12.

alfert:
#12 is handled in my current branch...

evnu:
Should I re-add the `notifications` configuration?

alfert:
No, thanks. Currently, there is no need for that.

PR #16, opened 2017-08-17
Bump PropEr to 1.2.0

evnu:
This bumps the PropEr dependency to 1.2.0. `mix test` did succeed on my machine (with Elixir v1.5) and on Travis for Elixir v1.4 and v1.5, but it failed on Travis with Elixir v1.3. If PropCheck should support Elixir v1.3, this needs more investigation.


evnu:
For the record, [this](https://travis-ci.org/evnu/propcheck/jobs/265708293) was the failing test run on Elixir v1.3.

evnu:
I believe the failing test mention above is caused by a race condition addressed in #17.

alfert:
Thanks again! ❤️

Issue #10, opened 2016-07-22
Store counterexamples on disk between test runs


alfert:
Any failure should not only print counterexamples, but also store them on disk. A test run should than search whether such a counterexample exists and use it, until the counterexample works or the file is deleted.

evnu:
I sometimes retrieve counter examples and manually store them in VCS to add explicit test cases. I would like the option to tell `property/4` where to store a counter example, if it finds one. Would this be in the scope of this issue? An additional option along the lines of `counterexample_path: <somepath>` could be used to tell that the user wants counterexamples to be stored on disk.

alfert:
Yes, this idea would be in the scope of this issue. 

My initial idea was to build something like Quvics Quickckeck, which stores some data (I assume the random seed since it is very small) in a file in the project directory. If this file exists, then only the counterexample are executed until they work and the file is deleted. Your approach is more generic, which is fine. I assume that we need to modify `property/4` to call `PropCheck.counterexample` instead of `quick check` and we must find a way to tell `property` to check for counter examples.

evnu:
> I assume that we need to modify property/4 to call PropCheck.counterexample

According to the [documentation](http://proper.softlab.ntua.gr/doc/proper.html#check-3), this should probably be a call to `PropCheck.check/3`. `PropCheck.counterexample/0` returns a counterexample, while `PropCheck.check/3` re-runs a counterexample.

alfert:
This looks like the things to do: 

* Create an ETS store for counterexamples, keyed by property name
* Store / Load the counterexamples during start/stop via :ets.tab2file
* property checks for a counterexample and runs it instead of the property
* extract property and counterexample handling from the macro for better testing and less macro magic
* provide a switch for mix to only run the counterexamples (true by default)
* provide a switch for mix where to store the counterexamples

evnu:
Two things that came to my mind while I was considering how to handle this:

1. If a property is renamed, pre-existing counterexamples might not be found anymore.
2. Also, the counterexamples cache needs to be cleaned from non-existing properties.

alfert:
#19 provides integration with mix and configuration options. Some docs are missing and will be the final step of implementation.

PR #18, opened 2017-08-28
Feature/persistent counter examples

alfert:
As discussed in #10 this is an implementation of persisting counter examples between test runs. The basic idea is to have a manager of counter examples called `CounterStrike` which stores all found counter examples as DETS on disk. 

When running a property, it is checked whether counter examples exists and if yes, then only the counter examples are executed and no other properties. This helps to focus on solving the failing properties with the same counter example to ease debugging.


alfert:
Yes, that would be the case. You could however save the file somewhere else and reuse it later.

The integration of CounterStrike into the property macro has to be done, yet.

Von unterwegs  gesendet

PR #25, opened 2017-10-02
Use defaults in produce

evnu:
Prior to this change, omitting the seed resulted in a badmatch with :undefined.

Note that this commit removes the helper functions and only wraps `:proper_gen.pick/3`. This simplifies the implementation.


alfert:
This looks good. Thanks for finding and removing the bug!

PR #27, opened 2017-10-11
Fix PropCheck.equals/2

evnu:
`Kernel.inspect/2` allows using options. For `:pretty`, a boolean value
must follow (`pretty: true`). Otherwise, an error similar to the
following is raised:

    Doctest for `:not_ok` fails with no function clause matching in Kernel.inspect/2.

This commit fixes PropCheck.equals/2 and adds tests (doctest and a test
to check that output is written to stdout on error).


alfert:
Ts ts ts, what testing all reveals :-)

Thanks for identifying and fixing the bug.

PR #29, opened 2017-11-29
Property documentation and test

evnu:
This changeset adds a little bit of documentation and a simple context test for the `property` macro and sets the README as the main page of the documentation. The reason for adding "context" to the documentation is to make the third parameter of `property` more obvious.


alfert:
Thanks!

PR #31, opened 2017-12-01
Fix typo in comment

pallix:


alfert:
Thanks for your PR. Your PR failed for an unknown reason and your well observed typo is also fixed in #29.

PR #33, opened 2018-01-22
Fix outdented heredoc warning on Elixir 1.6

evnu:
On Elixir 1.6, the following warning is written before this change:

    outdented heredoc line. The contents inside the heredoc should be
    indented at the same level as the closing """.


alfert:
Thanks!

PR #34, opened 2018-01-24
Stabilises State-Machine-Tests and enable Elixir 1.6 on Travis

alfert:
Originally started as documentation, this piece of work find a nasty flaw in the state machine implementation of the ping-pong master. It should now run stable.

### DIFF
--- a/test/support/movie_server.ex
+++ b/test/support/movie_server.ex
@@ -16,7 +16,13 @@ defmodule PropCheck.Test.MovieServer do
   def start_link(), do:
     GenServer.start_link(__MODULE__, [], name: __MODULE__)
 
-  def stop(), do: GenServer.call(__MODULE__, :stop)
+  def stop() do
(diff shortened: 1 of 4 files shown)


test/support/movie_server.ex
```diff
@@ -16,7 +16,13 @@ defmodule PropCheck.Test.MovieServer do
   def start_link(), do:
     GenServer.start_link(__MODULE__, [], name: __MODULE__)
 
-  def stop(), do: GenServer.call(__MODULE__, :stop)
+  def stop() do
```
evnu: Note that this function could also call `GenServer.stop/1`, which is synchronous. This is what's done in [movie_server.erl](http://proper.softlab.ntua.gr/code/movies/movie_server.erl) provides `movie_server:stop/0`.

alfert: Good Point. I will change both implementations, it is much more to the point.

PR #35, opened 2018-01-29
Run quickcheck after counterexample was successful

evnu:
Currently, when a counterexample (#10) was found and it did not fail, no further tests are performed until the test is re-executed at a later point in time. This may come as a surprise to users, as they probably expect that the property is checked completely on each run. This changeset changes this behaviour by running quickcheck if the counterexample did not fail the test.

Note that this might not capture the intend of #10, so feel free to close this PR if the current behaviour is the intended behaviour.

### DIFF
--- a/lib/properties.ex
+++ b/lib/properties.ex
@@ -85,12 +85,12 @@ defmodule PropCheck.Properties do
       :others ->
         # since the tag is set, we execute everything. You can limit
         # the amount of checks by using either --stale or --only failing_prop
-        PropCheck.quickcheck(p, [:long_result] ++opts)
+        qc(p, opts)
       {:ok, counter_example} ->
         # Logger.debug "Found counter example #{inspect counter_example}"
         result = PropCheck.check(p, counter_example, [:long_result] ++opts)
         with true <- result do
-            PropCheck.quickcheck(p, [:long_result] ++ opts)
+          qc(p, opts)
         else
           false -> {:rerun_failed, counter_example}
           e = {:error, _} -> e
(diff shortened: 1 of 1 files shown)


alfert:
That's an interesting idea. The very basic idea of #10 is to focus on the failing property and expects that you work on that property only until it works (which is good, but you cannot guarantee that way of working, of course. At least, I do not follow it in all consequence). In `StreamData`, they use the ability of `ExUnit` to infer which tests must be rerun because some called modules are modified after the last test.  
Your idea tests all properties automatically, after the failing property succeeds again. This also ensures that all other modified implementation are checked. 

What do you think is a more usable approach and why? These features are all about usability and I am not sure what is best - as a default and as a configurable command line option.

evnu:
> What do you think is a more usable approach and why?

I know that I stumbled over forgetting to re-run the property after fixing a failing example. Example: I fixed the counter example and proceeded with running the complete test suite again, only to encounter another example failing with that property. At that point in time, I thought I had already fixed the property and forgot that it was only a single case which was fixed.

I do believe that both approaches have merit. Usually when fixing a property, I am focused on fixing only this one item and only run it. So the expectation that I am working only on that property usually applies.

I think I would be fine with a configuration option. Adding a command line option would work and allows adding a section to the README indicating how counterexamples work and how to change the behaviour. Additionally, when running a counterexample successfully to completion, a message could be written indicating that the property should be re-run to get a complete sample, or to use the command line flag in the future if that is what the user would prefer.

evnu:
@alfert Did you have time to consider this PR again? I again stumbled over this problem today.

alfert:
Let's go for it! But we need some documentation of this behaviour and the conflict must be resolved (besides a rebase, of course). Could you please take care of this? I will merge then.

evnu:
I will take care of it.

evnu:
I rebased, introduced a small helper function to refactor the identical calls of `PropCheck.quickcheck/2`, and added a paragraph to the property documentation in 21ecd52. Is the documentation enough?

alfert:
Looks good, thanks! I will publish a release today.

PR #37, opened 2018-02-02
Handling errors

### LINKED ISSUE #36: Handling PropEr:quickcheck/2 error values
Currently, `{:error, }` from `PropEr:quickcheck/2` is considered a counterexample (#10) by PropCheck. This leads to unexpected behaviour.

For example, re-running a property after it failed due to `:cant_generate` ends with an unhelpful `:non_boolean_result` message. Consider the following example:

```elixir
defmodule FailingTest do
  use ExUnit.Case
  use PropCheck

  property "failing" do
    gen = such_that b <- false, when: b # will produce {:error, :cant_generate}
    forall b <- gen do
      b
    end
  end
end
```

First run ends as expected:

```
  1) property failing (FailingTest)
     test/failing_test.exs:5
     Property Elixir.FailingTest.property failing() failed. Counter-Example is:
     {:error, :cant_generate}
     
     code: nil
     stacktrace:
       (propcheck) lib/properties.ex:109: PropCheck.Properties.handle_check_results/3
       test/failing_test.exs:5: (test)
```

The second run, OTOH, ends with `:non_boolean_result`:

```
  1) property failing (FailingTest)
     test/failing_test.exs:5
     Property Elixir.FailingTest.property failing() failed. Counter-Example is:
     {:error, :non_boolean_result}
     
     code: nil
     stacktrace:
       (propcheck) lib/properties.ex:109: PropCheck.Properties.handle_check_results/3
       test/failing_test.exs:5: (test)
```

Re-running the property can only be done by removing the counterexamples file, which is suboptimal.

The error handling in https://github.com/alfert/propcheck/blob/master/lib/properties.ex#L107 must be improved. I think this should be `counterexample when is_list(counterexample)`.

In a similar fashion, https://github.com/alfert/propcheck/blob/master/lib/properties.ex#L106 might not do what is expected. `should_fail` should probably still output in error if no type instances were generated.

@alfert I can come up with a PR, if you are interested.

evnu:
Fix #36. The PR contains a simple test case as well (a78787c).

### DIFF
--- a/lib/properties.ex
+++ b/lib/properties.ex
@@ -97,14 +97,20 @@ defmodule PropCheck.Properties do
   # In this method a new found counter example is added to `CounterStrike`.
   defp handle_check_results(results, name, should_fail) do
     case results do
+      error = {:error, _} ->
+        raise ExUnit.AssertionError, [
+          message:
+            "Property #{mfa_to_string name} failed with an error: #{inspect(error)}",
+          expr: nil
+        ]
       true when not should_fail -> true
       true when should_fail ->
         raise ExUnit.AssertionError, [
           message:
             "Property #{mfa_to_string name} should fail, but succeeded for all test data :-(",
           expr: nil]
-      _counter_example when should_fail -> true
-      counter_example ->
+      counter_example when is_list(counter_example) and should_fail -> true
+      counter_example when is_list(counter_example) ->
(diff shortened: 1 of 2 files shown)


alfert:
Looks good. I will test and merge after my vacation.

PR #38, opened 2018-02-21
List available mix tasks

evnu:
By calling `mix propcheck`, the available mix tasks are listed:

    $ mix propcheck
    Available PropCheck tasks:

    mix propcheck.clean   # Removes the counter example file of propcheck
    mix propcheck.inspect # Inspects and prints all counter examples.

Additionally, the moduledoc is now used to generate the output for
`mix help propcheck`.

### DIFF
--- a/lib/mix.ex
+++ b/lib/mix.ex
@@ -13,6 +13,9 @@ defmodule PropCheck.Mix do
 end
 
 defmodule Mix.Tasks.Propcheck do
+  use Mix.Task
+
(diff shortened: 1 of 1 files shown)


evnu:
It seems that the mix tasks are not found when `PropCheck` is used as a dependency:

```
$ mix propcheck.clean     
** (Mix) The task "propcheck.clean" could not be found
```

alfert:
Thanks again!

PR #39, opened 2018-03-05
Fix bitstring/1 documentation

evnu:
Fix 14c4d21d. bitstring/1 needs a bit size of Len.


evnu:
For reference, see also [`proper_types:bitstring/1`](http://proper.softlab.ntua.gr/doc/proper_types.html#bitstring-1):

> All bitstrings with a bit size of Len. 

(@jvf found the error in the documentation)

alfert:
what a stupid mistake. Thanks again!

PR #44, opened 2018-03-20
Remove control characters from doc strings

evnu:
This removes backspace (`^H` in vim) control characters from doc strings.


alfert:
Thanks!

PR #47, opened 2018-04-23
Allows more than two bindings in a let

adkron:
When cycling through the let with the `,` in the list we don't append
the list as expected. This works fine with only two items in the
bindings list, but with more than that we start to get nested lists
which turn into argument errors when we try to call `elem(x, 0)`.
Changing to a pipe fixes the issue because it will concatenate the
lists, but the final recursion needs to return a list to be appended
too.

### DIFF
--- a/lib/propcheck.ex
+++ b/lib/propcheck.ex
@@ -541,10 +541,10 @@ defmodule PropCheck do
     end
     defp let_bind({:<-, _, [var, rawtype]} = _bind), do: {var, rawtype}
     defp let_bind([{:<-, _, [var, rawtype]}]) do
-      {var, rawtype}
+      [{var, rawtype}]
     end
     defp let_bind([{:<-, _, [var, rawtype]} | rest]) do
-      [{var, rawtype}, let_bind(rest)]
+      [{var, rawtype} | let_bind(rest)]
     end
 
     @doc """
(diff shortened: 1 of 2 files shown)


adkron:
When generating larger structures the positional arguments of the lists get extremely difficult to follow. That is also not how the documentation shows it. I did that as a workaround, but I’m now pointing at a fork.

Thank you for propcheck by the way. We mentioned it in the first two episodes of Elixir  Outlaws. Elixiroutlaws.com

Sorry if I pronounced your name wrong. Episode 2 is all about property testing. 

Amos King
Binary Noggin

alfert:
@adkron, Thanks for your friendly words in the podcast. Your are right that documentation should be more user-friendly. I am also thinking about adding a new macro-based DSL to stateful-testing as discussed in the context of StreamData.

What a nasty bug and even more a nice example of insufficient testing. Using `PropCheck` for testing `PropCheck` would have been nice, but where is the fix point 😃 ?

adkron:
I tried to think about how to test propcheck with propcheck but I couldn’t quite figure that out. I could have generated an AST and passed it through the function the macro uses but I really wanted to go through the macro. Honestly I wasn’t super happy with the test but it was better than me not putting one at all. Thanks for all the work you’ve done. 

Amos King
Binary Noggin

PR #48, opened 2018-04-26
Handle assertions in forall macro

evnu:
Users often attempt to use ExUnit assertions within the forall macro:

    use PropCheck
    property "assertion" do
        forall n <- nat() do
            assert n < 0
        end
    end

In :quiet mode, no indication that an assertion failed is written. In
verbose mode, the exception cannot be handled by PropEr itself:

    Failed: The input fails the test.
    An exception was raised: error:#{'__exception__'=>true,'__struct__'=> [...]
    Stacktrace: [{'Elixir.ExUnit.Assertions',assert,2,
                     [{file,"lib/ex_unit/assertions.ex"},{line,330}]}].

The exception message must be formatted manually by the user.

This commit changes this by wrapping the property within forall in a try
block. If an ExUnit.AssertionError is triggered, diagnostics similar to
the ones used in ExUnit are written:

    Assertion with < failed, both sides are exactly equal
    code: assert n < 0
    left: 0
    stacktrace:
        (ex_unit) lib/ex_unit/assertions.ex:330: ExUnit.Assertions.assert/2
        test/propcheck_test.exs:48: anonymous fn/1 in PropcheckTest."property assertion"/1
        (proper) src/proper.erl:1353: :proper.apply_args/3
        (proper) src/proper.erl:1040: :proper.retry/3
        (propcheck) lib/properties.ex:90: PropCheck.Properties.execute_property/3
        test/propcheck_test.exs:46: PropcheckTest."property assertion"/1
        (ex_unit) lib/ex_unit/runner.ex:306: ExUnit.Runner.exec_test/1
        (stdlib) timer.erl:166: :timer.tc/1
        (ex_unit) lib/ex_unit/runner.ex:245: anonymous fn/4 in ExUnit.Runner.spawn_test/3


evnu:
Note that a failed assertion always results in output, even if `:quiet` is set. I did not find a way to pass in the verbosity into the macro.

evnu:
CC @pallix

test/propcheck_test.exs
```diff
@@ -42,6 +42,16 @@ defmodule PropcheckTest do
     end) == "[1,2,3]\n"
   end
 
+  test "can use assertion in forall" do
+    use PropCheck
+    assert capture_io(fn ->
+      quickcheck(
+        forall _x <- :not_ok do
+          assert false
+        end)
```
alfert: you could add the `[:quit]` or `[:verbose]` as an option to `quickcheck`.

lib/propcheck.ex
```diff
@@ -341,7 +341,23 @@ defmodule PropCheck do
     defmacro forall(binding, property)
     defmacro forall({op, _, [var, rawtype]}, do: prop) when op in @in_ops do
```
alfert: What about adding an optional `option` parameter to `forall`, which is evaluated regarding verbosity? Since we are generating the environment for the property, we could handle this stuff here. Do I miss something?

alfert:
👍 Brilliant idea. Using ExUnit assertions in properties is the one feature I like most about StreamData. So providing it here gives PropCheck the ability to become more Elixirish. 

We should handle the log output nicely. Did you try adding options? See my comments on the code.

test/propcheck_test.exs
evnu: That option can currently not be passed down into `forall`, as it is not part of the property itself. `quickcheck` also delegates to PropEr directly, so if we want to inject an option into the property, we probably need to change that.

lib/propcheck.ex
evnu: I believe it would be simpler to add capture wrapping in the property macro. There, we already have the information regarding verbosity. This does not help with calling into `quickcheck` directly (see https://github.com/alfert/propcheck/pull/48#discussion_r184658101), but maybe creating output is ok for that other case.

If capturing in the property macro is not possible, we would need to inject code into the property passed into the macro. This will definitely make the implementation more complicated, but maybe it is possible.

evnu:
1b591c1 silences the `forall` output. This is ok as a workaround, but I would prefer it if we were able to pass the verbosity from the property itself down into that function.

alfert:
Thanks for that fix! 
Could you please do a rebase? After that I will the PR.

evnu:
I rebased the commits and squashed the two fixup commits into where they belong. For a future PR, it might also be useful to add similar functionality to other macros, such as `implies/2`.

PR #51, opened 2018-05-16
Fixes documentation titles

zamith:
Why:

* Markdown requires an empty line after the titles

This change addresses the need by:

* Adding that extra line across the project, which will affect how it
  rendered on hexdocs

### DIFF
--- a/lib/fsm.ex
+++ b/lib/fsm.ex
@@ -18,6 +18,7 @@ defmodule PropCheck.FSM do
   use the  <a href="#functions">API functions </a> of `PropCheck.FSM`.
 
   ## The states of the finite state machine
+
   Following the convention used in `gen_fsm behaviour`, the state is
   separated into types `t:state_name/0` and some
   `t:state_data/0`. `state_name` is used to denote a state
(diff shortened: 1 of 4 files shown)


alfert:
Thanks for identifying and correcting the bad formatted docs!

Your diff is way bigger than expected when reading the comments. Did you use the Elixir formatter for doing the fixes? There many edits all over the place, e.g. changing indentations, adding parentheses, adding line breaks for longer functions definitions, in sum these are more than 900 changed lines. Some of them are nice, some of them I do not like that much. For example, in `fsm.ex` there were no problems regarding headlines (at least at hexdocs.pm and in Dash, there are no problems).

zamith:
Ah, yes, my editor has the formatter running automatically. I can revert those.

The changes should be mostly the headings, except for the propcheck file itself, which I had to indent the entire documentation section in order for the lists to render properly.

zamith:
@alfert Ok, removed all the formatter changes.

alfert:
Perfect and Thanks!

PR #54, opened 2018-05-31
Feature DSL for stateful property testing

alfert:
Provides a DSL defining commands together with their argument generation, implementation, pre- and post condition and the next state function. It is inspired by the Quvics ECQ.


alfert:
@evnu, yes, the specs were wrong. Dialyse loves this version better.

PR #62, opened 2018-08-12
Corrects PropCheck.StateM.DSL.weight/1 documentation

adkron:
The documentation shows that there are two arguments to `PropCheck.StateM.DSL.weight/1` in the `moduledoc`. The documenation contradicts both the code and the documentation at the function level later. That leads me to believe the documentation was wrong, and not the implementation.

### DIFF
--- a/lib/statem_dsl.ex
+++ b/lib/statem_dsl.ex
@@ -134,10 +134,8 @@ defmodule PropCheck.StateM.DSL do
   in function `weight/2`, taking the current model state and the command to be
   generated. The return value is an integer defining the frequency. In our cache
   example we want three times more `find` than other commands:
-
-      def weight(_state, :find),  do: 1
-      def weight(_state, :cache), do: 3
-      def weight(_state, :flush), do: 1
+  
+      def weight(_state), do: %{find: 1, cache: 3, flush: 1}
 
   ## The property to test
   The property to test the stateful system is for all systems more or less


alfert:
@adkron thanks for pointing out this inconsistency!

Looking into it, I found that there are even more inconsistencies: the spec for the `weight` call is also wrong and the `moduledoc` links to a function and to a callback.

In your patch, `weight/1` returns a map, in my test examples (e.g. `counter_dsl.exs`) I return keyword lists. The code works with both of them thanks to `Enum`. Do you have a preference of using maps instead of a keyword list here and why? From a stylistic point of view I would prefer only one type of return values.

adkron:
I'm not sure which I prefer. Why prefer one over another?

I feel like the weights are options and should be keywords because of that. At the same time keywords have a concept of more than one item, and ease of merging them on top of each other. That creates a simple way to pile the configurations and use just the newest. I don't see us doing that here. Keyword lists are great when you need to preserve user's order. I don't think we need that here either.

A map feels more straightforward because it provides a single value for each key. Maps are faster to access since they aren't a list that has to be traversed to find the data. Maps allow pattern matching, and keyword lists cannot.

I think I've convinced myself to use maps. What do you think?

alfert:
Your reasoning is valid. And since the frequency generator (which is fuelled by the weights) takes a list of frequency/generators pairs, a reconstruction is required anyway. So let's go for maps.

I do the other changes as well.

PR #67, opened 2018-08-23
Fix typos in statem_dsl

evnu:


alfert:
I produced an  impressive amount of typos. Thanks for fixing!

PR #72, opened 2018-10-15
Adding in map support which requires 1.3 proper

IRog:
I also removed the `-dev` from the version, which hopefully isn't a problem.
Let me know, if you want any changes.

### DIFF
--- a/lib/basic_types.ex
+++ b/lib/basic_types.ex
@@ -104,6 +104,12 @@ defmodule PropCheck.BasicTypes do
 
 
   @doc """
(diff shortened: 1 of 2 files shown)


alfert:
Thank you! But please revert the removal of the `-dev`, since it is crucial for my automated release process. Background: This tag notes that the generated version is not the final one, but an intermediate non-releasable state. It helps in particular to prevent accidental releases. Read the version as "this is the development towards 1.1.3". 

After the removal I will apply your PR.

IRog:
@alfert removed. Thanks!

Issue #73, opened 2018-10-24
** (Mix) The task "propcheck.clean" could not be found


building39:
`mix.exs`:

```defmodule Pbt.MixProject do
  use Mix.Project

  def project do
    [
      app: :pbt,
      version: "0.1.0",
      elixir: "~> 1.7",
      start_permanent: Mix.env() == :prod,
      deps: deps()
    ]
  end

  # Run "mix help compile.app" to learn about applications.
  def application do
    [
      extra_applications: [:logger]
    ]
  end

  # Run "mix help deps" to learn about dependencies.
  defp deps do
    [
      {:propcheck, "~> 1.0", only: :test}
    ]
  end
end```

Run `mix propcheck.clean`:
```$ mix propcheck.clean
Compiling 1 file (.ex)
Generated pbt app
** (Mix) The task "propcheck.clean" could not be found```

**Comment by evnu** [comment]
The task cannot be find as `:propcheck` is only available for `MIX_ENV=test`. As a workaround, this should work:

```
$ MIX_ENV=test mix propcheck.clean
```

evnu:
Note that you can also use `only: [:test, :dev]`, then `MIX_ENV=test` is not necessary.

@alfert I usually allow propcheck in `MIX_ENV=dev` in order to use generators in `iex`. Maybe the README should specify `only: [:test, :dev]`?

alfert:
Yeah, this is a good idea, indeed!   An example with an IEX session would also be nice.

Von unterwegs  gesendet

building39:
Yeah, I shoulda thought of that before opening this issue. You can close it.

PR #74, opened 2018-10-26
Enhance message on purging counterexamples

evnu:
This enhances the message introduced with 4692ef8, which outputs an info message on purging counterexamples. This should help avoid users stumbling over `mix propcheck.clean` missing in `MIX_ENV=dev` if propcheck is not allowed in `:dev` (see #73).


alfert:
Thanks again!

Issue #70, opened 2018-10-04
Support Targeted Property-Based Testing


zkessin:
We should work to include the new features of PropER 1.3

alfert:
Correct. I have an WIP feature branch already in my local repo, which I should publish. I would like to get PRs on that.

zkessin:
I will be happy to help out

alfert:
@zkessin, I published my current state of work on this on its own branch and I am open for reviews, PRs and the like. Currently the tests fail. Any hints why and what to do are more than welcome.

zkessin:
Awesome, I will take a look in the next few days if I have a chance

alfert:
I update the branch. What works now is the level example with `forall_sa` and `forall_targeted`. We are in now in sync with the tutorial from Proper in both variants, web page and source code, which interestingly differ. Stil falling is a level implementation based on `exists` or `not_exists`. 

Generally, I am struggling with the entire concept. The idea of having properties with a negative logic (if they fail, everything is fine) irritates me. And to find a useful metric expressed in a real number to optimise for and which is related with the correctness of software in general, is by no means obvious and I doubt if this is generally a valid approach. It reminds to program synthesis with genetic algorithms in the 90ies: is a sorting algorithm which sorts in 95% of all cases correctly better than an algorithm with sorts only in 94% of all cases? So, what to optimise for? But it might be that I look from the wrong point of view on this approach. If anybody can explain in which cases the targeted approach helps, I would be delighted.

alfert:
Just to copy the final comment from #77: 

> This was a quite a tour. The biggest issue - beside understanding and valuing the entire approach - was is little copy bug that changed the semantics of exists and not_exists. So it is finally done.

> Documentation can be improved tremendously, but currently I would suggest to read Fred Hebert's book Property based testing, which include a chapter on targeted PBT.

It it available in the master now, a new release hex is coming soon.

PR #77, opened 2018-11-10
Supporting Targeted Property Based Testing

alfert:
As requested in #70, here are the macros and docs for targeted property based testing as implemented in PropEr 1.3

Documentation is more or less as in PropEr, so it does not really help.


alfert:
This was a quite a tour. The biggest issue - beside understanding and valuing the entire approach - was is little copy bug that changed the semantics of `exists` and `not_exists`. So it is finally done. 

Documentation can be improved tremendously, but currently I would suggest to read Fred Hebert's book Property based testing, which include a chapter on targeted PBT.

Issue #82, opened 2019-02-04
Bug in level test?


evnu:
As can be seen in the [tests of PropCheck](https://github.com/alfert/propcheck/commit/4af8b8804874e855c7124679ac56d081bb0aadf1#diff-f0234345bd20e26cbfd8f03f6135a08fR36), we encounter a `{$used, _, _}` tuple with TPBT. I followed the [tutorial](http://proper.softlab.ntua.gr/Tutorials/PropEr_testing_with_search_strategies.html) to get a first glimpse into PropEr's TPBT, where I stumbled over this as well. It seems unclear how to correctly interpret this internal data. In PropCheck's test, the interpretation seems to be `{:"$used", old, new}`. I think that the interpretation should be `{:"$used", new, old}`.

In my implementation, I only generate a single additional step for each invocation of `path_next_sa()`:

```elixir
  def path_next_sa() do
    fn
      old_path, _temperature when is_list(old_path) ->
        let next_step <- step() do
          old_path ++ [next_step]
        end

      {:"$used", _new_step, old_path}, _temperature when is_list(old_path) ->
        let next_step <- step() do
          old_path ++ [next_step]
        end
    end
  end
```

If I adapt PropCheck's test to also generate a single new step, while maintaining that the old path is the second tuple element, this fails.

```elixir
  def path_next_sa() do
    fn
      old_path, _temperature when is_list(old_path) ->
        let next_step <- step() do
          old_path ++ [next_step]
        end

      {:"$used", old_path, _next_step}, _temperature when is_list(old_path) ->
        let next_step <- step() do
          old_path ++ [next_step]
        end
    end
  end
```

Error:

```
** (FunctionClauseError) no function clause matching in anonymous fn/2 in MazeGameTest.path_next/0

     The following arguments were given to anonymous fn/2 in MazeGameTest.path_next/0:
     
         # 1
         {:"$used", :down, [:right, :down]}
     
         # 2
         1.0
```

alfert:
Hmm, I am puzzled, to be honest. My implementation in `level_tpbt.exs` works (but only on my machine and not on travis which I do not understand - but this is a different story). 

```elixir
def path_next_sa() do
    fn
      (prev_path, _temperature) when is_list(prev_path) ->
          let next_steps <- vector(20, step()), do:
            prev_path ++ next_steps
      ({:"$used",  prev_path, _another_path}, _temperature) when is_list(prev_path) ->
       let next_steps <- vector(20, step()), do:
[...]

PR #86, opened 2019-02-07
Fix level test

evnu:
This is an attempt at fixing two issues described in #82:

1. Avoiding the PropEr-internal `{:"$user, _, _}` tuple
1. Avoiding to use legacy TPBT macros

The first item is resolved by the second item. It seems that when we use the legacy macros, the internal tuple is returned to the neighbourhood function. When using `forall_targeted` instead of `forall_sa`, this is not longer the case.

Note that when using `forall_targeted`, `collect/1`, can currently not be used, as the macro attempts to apply `:erlang.not/1` onto the result of `collect/1`. Additionally, when using a proper-derived NF, we need to perform more tests until a path is found, as the search space is apparently traversed more randomly.

### DIFF
--- a/test/level_tpbt_test.exs
+++ b/test/level_tpbt_test.exs
@@ -33,9 +33,6 @@ defmodule PropCheck.Test.LevelTest do
       (prev_path, _temperature) when is_list(prev_path) ->
           let next_steps <- vector(20, step()), do:
             prev_path ++ next_steps
-      ({:"$used",  prev_path, _another_path}, _temperature) when is_list(prev_path) ->
-       let next_steps <- vector(20, step()), do:
-          prev_path ++ next_steps
     end
   end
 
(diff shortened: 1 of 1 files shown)


evnu:
@alfert I would have expected travis to run on this PR. Did that happen?

evnu:
Note that this can be closed if another solution is preferable. I only wanted to document what I did to understand what is going on with TPBT.

alfert:
@evnu, the travis run failed miserably due to the setup problems (but not in 1.6 - I am surprised again) 

https://travis-ci.org/alfert/propcheck/builds/490113712

Nevertheless, your approach seems right. It does not make sense to fight against windmills just for the sake of it.

alfert:
I merge this, because it makes simply sense. I will rebase my approach on top of this, so that we come to a clean solution.

evnu:
@alfert Do you know why travis is not linked from the commits in this PR?

alfert:
I assume due to the conversion from Travis.org to .com

Another issue to look into 

Von unterwegs  gesendet

PR #94, opened 2019-03-18
Update dependencies to get rid of compile warnings

evnu:
Updated with `mix deps.update --all`:

* certifi 2.0.0 => 2.5.1
* coverex 1.4.15 => 1.5.0
* ex_doc  0.19.2 => 0.19.3
* hackney 1.9.0 => 1.15.1
* idna    5.1.0 => 6.0.0
* mimerl  1.0.2 => 1.2.0
* poison  3.1.0 => 4.0.1
* ssl_verify_fun 1.1.1 => 1.1.4
* unicode_util_compat 0.3.1 => 0.4.1

`poison` breaks CI, as it requires Elixir `~> 1.6`. @alfert Do you prefer to maintain `PropCheck` for older Elixir versions as well? I would then revert the change to `poison`.


evnu:
In case that CI is not reported, [here](https://travis-ci.org/evnu/propcheck/jobs/508085302 is the failing build. Error:

```
warning: the dependency :poison requires Elixir "~> 1.6" but you are running on v1.5.3
Compiling 4 files (.ex)
== Compilation error in file lib/poison/parser.ex ==
** (CompileError) lib/poison/parser.ex:276: undefined function defguardp/1
```

The [other builds](https://travis-ci.org/evnu/propcheck/builds/508085301) did not fail

alfert:
Since I just updated the version requirements, I would currently stick to 1.5. Generally, I would like to be conservative in the Elixir requirements to support more projects - independent from missing knowledge which projects use PropCheck and rely on earlier versions of Elixir. 

So, yes, please update the dependencies. Poison comes from coverex, requiring a 1.4.15 for Elixir 1.5. Seems that version hell finally arrives are Elixir :-/ not really surprising...

evnu:
Reverted `poison` with 2f1eb65 by locking coverex to only update on patch releases.

PR #100, opened 2019-03-20
Lint with Credo

### LINKED ISSUE #98: Consider using Linter
I think the code base and CI might benefit from a linter. I found [credo](https://github.com/rrrene/credo) useful in other projects. If there is interest for this, I will put up a PR.

evnu:
This pull requests adds [credo](https://github.com/rrrene/credo) to lint `PropCheck`. This lint is also done by CI.

See [here](https://circleci.com/gh/evnu/propcheck/58) for the initial list of things Credo complains about when using `mix credo --strict`. I added a configuration with `mix credo gen.config`.

Fix #98.

### DIFF
--- a/lib/result.ex
+++ b/lib/result.ex
@@ -9,7 +9,7 @@ defmodule PropCheck.Result do
   @type t :: %__MODULE__{tests: [any], errors: [any], current: nil | atom}
 
   def start_link do
-    :gen_server.start_link({ :local, __MODULE__ }, __MODULE__, [], [])
+    :gen_server.start_link({:local, __MODULE__}, __MODULE__, [], [])
   end
   def stop do
     :gen_server.call(__MODULE__, :stop)
@@ -26,7 +26,7 @@ defmodule PropCheck.Result do
 
  @spec init(any) :: {:ok, t}
  def init(_args) do
-    { :ok, %__MODULE__{} }
+    {:ok, %__MODULE__{}}
   end
 
   @spec handle_call({:message, any, any}, {pid, any}, t) :: {:reply, :ok, t}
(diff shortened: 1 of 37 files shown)


evnu:
I will add a commit for each type of warning credo writes out. @alfert you can then decide which of those things you find useful, and which to disable warnings to disable with Credo's configuration. Does that work for you?

evnu:
I fixed all messages by Credo, except the warning that a pipe should begin with a raw value. That warning is disabled with 33c2271. If you prefer it the way credo suggests, I revert that commit and fix those warnings. For example, by default, credo complains about something like this:

```elixir
function1(value) |> function2()
```

Instead, it prefers the upper like this:

```
value |> function1 |> function2
```

evnu:
I think it would probably be worthwhile to squash all commits prefixed with `lint:` into a single commit after a decision which warnings to actually obey has been made.

lib/result.ex
```diff
@@ -12,11 +12,9 @@ defmodule PropCheck.Result do
     :gen_server.start_link({ :local, __MODULE__ }, __MODULE__, [], [])
   end
   def stop do
-    try do
-      :gen_server.call(__MODULE__, :stop)
-    catch
-      _ -> :ok
-    end
+    :gen_server.call(__MODULE__, :stop)
```
alfert: I am not sure that I like this and to be honest I didn't know that this possible at all. For a minimal function like this is perfectly ok, but in a larger context I would like to know what the scope of the try-catch block. If this rule only applies to entire functions, then I am fine.

alfert: I checked https://github.com/rrrene/credo/issues/224 and the Elixir docs. So it only makes sense for a function that starts with a try block closes with the corresponding catch etc block. So, I think this is ok.

alfert:
Great, thank you for this tedious work 💚 

I really like these little commits which a clear change, so no need for squashing.

evnu:
Thanks for merging!

Note that implicit try in functions is a controversial feature. See the [proposal](https://github.com/rrrene/credo/issues/224). I don't prefer it either instead of explicit `try..rescue..catch..after..end`.

Issue #101, opened 2019-03-21
Tests killed/crashed with segfault on CI


evnu:
I found two test runs which failed in bad ways without a lot of data to debug the issue. Interestingly, both cases here were hit within the `property A left-heavy tree`.

[Build 10](https://circleci.com/gh/alfert/propcheck/10), Elixir v1.5

```
* property A left-heavy tree
[...]
Received 'killed' signal
```

[Build 82](https://circleci.com/gh/evnu/propcheck/82), Elixir v1.7

```
* property A left-heavy tree
[...]
Crash dump is being written to: erl_crash.dump...Received 'segmentation fault' signal
```

There is nothing which can be done right now, but this is at least curious.

It could be worthwhile to upload core dumps as artifacts in CircleCI. That might enable some debugging, e.g. with [recon](https://github.com/ferd/recon/blob/master/script/erl_crashdump_analyzer.sh).

evnu:
See the [CircleCI](https://circleci.com/docs/2.0/artifacts/#uploading-core-files) documentation on how to upload core files.

evnu:
With the PR #120, we no longer store crashes as we would need to upload them to S3 when using travis.

evnu:
This did not happen for a long time, I think we can close this.

PR #102, opened 2019-03-21
ci: Store core and crash dumps

evnu:
Part of #101. Enable storing crash and core dumps in CircleCI. See [the documentation](https://circleci.com/docs/2.0/artifacts/#uploading-core-files) on how to retrieve such artifacts, if any core or crash dumps were written.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,6 +17,7 @@ jobs:
           keys:
             - testing-elixir-v<< parameters.version >>
       - run: MIX_ENV=test mix do deps.get, compile
+      - run: ulimit -c unlimited
       - run: mix test --cover --trace --exclude will_fail:true --exclude unstable_test:true
       - store_test_results:
           path: _build/test/lib/propcheck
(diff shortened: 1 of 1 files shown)


evnu:
@alfert Configuring CircleCI is a bit cumbersome, as the YAML configuration is easy to get wrong. I found [validating the config locally](https://circleci.com/blog/circleci-hacks-validate-circleci-config-on-every-commit-with-a-git-hook/) to be useful.

alfert:
This is a sweet bug 🤕  : The build fails if no erlang crash occurs. Build automation can be complicated ...
And yes: The idea of locally checking is great except that you also need to have a tool from CircleCI locally installed, if I understand it correctly.

evnu:
> The idea of locally checking is great except that you also need to have a tool from CircleCI locally installed, if I understand it correctly.

I downloaded the release directly from [github](https://github.com/CircleCI-Public/circleci-cli/releases), unpacked it and ran the tool directly from that directory. The `wget | sh` command is not needed.

evnu:
I also tried to inject a failure to check if CircleCI will try to attempt to copy the crash dumps. See [here](https://circleci.com/gh/evnu/propcheck/109) for such an experiment. That run has an additional step (`mkdir`,..), where CircleCI attempts to copy `erl_crash.dump` and `core.*`. So hopefully this will store dumps on CI the next time it happens.

evnu:
@alfert Anything more to do here?

alfert:
No, sorry, just too much other things. Thanks for this improvement!

PR #106, opened 2019-04-10
Configure storing counter examples with tags

### LINKED ISSUE #105: Disable persisting counter examples for specific tests
I have a generator which takes its allowed values from a database. For each invocation of `mix test`, a migration is run where such values are inserted into a database. After the test, the database is rolled back. This results in different row ids over subsequent test runs. Now, if a counter example is found, the stored counter example will not match any stored entry in the database on the next run. I would like to explicitly mark a property to not store the counter example upon error:

```elixir
@tag counter_example: false
property "do not store counter example" do
  ...
end
```

evnu:
This tag allows to configure that counter examples should not be stored. The tag can be used in `@moduletag`, `@describetag` and a regular `@tag`. It can also be used multiple times with the given precedence: tag > describetag > moduletag

The message of a failed property indicates if a counter example was stored or not. For example, if storing a counter example is disabled by any of the tags mentioned above, the following is appended to the error:

> Counter example NOT stored, :store_counter_example is set.

Note that testing this feature is cumbersome. I added `test/store_counter_examples_test.exs` to check that it works, but apparently there is no way to tell ExUnit to ignore some intentionally failing test cases. The tests use an `on_exit` handler registered with `setup_all` in order to ensure that the counter examples are checked after the test cases are run.

See [this](https://circleci.com/gh/evnu/propcheck/146) pipeline, were I ran the tests explicitly to check that counter examples are not stored. Example output:

```
1) property failing (StoreCounter.ExampleTag)
     test/store_counter_examples_test.exs:63
     Property Elixir.StoreCounter.ExampleTag.property failing() failed. Counter-Example is:
     [0]
     
     Counter example NOT stored, :store_counter_example is set.
     
     code: nil
     stacktrace:
       (propcheck) lib/properties.ex:185: PropCheck.Properties.handle_check_results/4
       test/store_counter_examples_test.exs:63: (test)
```

@alfert Maybe you have an idea how to test this in CI. The only option I currently see is to store the expected output as a file, and compare the output of `mix test test/store_counter_examples_test.exs` with the expected output.

Fix #105.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,6 +19,10 @@ jobs:
       - run: MIX_ENV=test mix do deps.get, compile
       - run: ulimit -c unlimited
       - run: mix test --cover --trace --exclude will_fail:true --exclude unstable_test:true
+      - run:
+          command: |
+            mix test test/store_counter_examples_test.exs --include will_fail:true && exit 1
+            mix test test/verify_counter_examples_test.exs
       - store_test_results:
           path: _build/test/lib/propcheck
       - save_cache:
--- a/lib/properties.ex
+++ b/lib/properties.ex
@@ -179,7 +179,7 @@ defmodule PropCheck.Properties do
             CounterStrike.add_counter_example(name, counter_example)
             "Counter example stored."
           else
-            "Counter example NOT stored, :store_counter_example is set."
+            "Counter example NOT stored, :store_counter_example is set to false."
           end
 
         raise ExUnit.AssertionError, [
--- a/lib/properties.ex
+++ b/lib/properties.ex
@@ -39,6 +39,47 @@ defmodule PropCheck.Properties do
   After a property was ran successfully against a previous counter example, PropCheck will
   run the property again to check if other counter examples can be found.
 
+  ### Disable Storing Counter Examples
+
+  Storing counter examples can be disabled using the `:store_counter_example` tag. This
+  can be done in three different scopes: module-wide scope, describe-wide scope or for
+  a single property.
+
+  **NOTE** that this facility is meant for properties which cannot run with a value generated
+  in a previous test run. This should usually not be the case, and `:store_counter_example`
+  should only be used after careful consideration.
+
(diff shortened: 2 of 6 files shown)


alfert:
Thanks for the PR!

I would suggest to change the message to include the boolean value, i.e. 

    Counter example NOT stored, tag :store_counter_example is set to false

Otherwise it sounds strange that `store_...` is set and thus the example will *not* be stored. 

Regarding testing: Since we can examine the stored counter examples, we can run after the initial test as in your pipeline, a second test run, where we check that there are no counter examples stored for these test cases. The two test runs can be combined in a separate mix command by an alias as we done for the `clean` alias: 

    mix test test/store_counter_examples_test.exs
    mix test test/verify_counter_examples_test.exs

What doing you think?

evnu:
> I would suggest to change the message to include the boolean value, i.e.

Yes, that makes sense.

> Regarding testing: Since we can examine the stored counter examples, we can run after the initial test as in your pipeline, a second test run, where we check that there are no counter examples stored for these test cases.

I was thinking about that as well. I did not want to clutter the test pipeline, but if you are ok with that, this seems the way to go. It feels weird to have such a one-off test for a single feature, but I cannot think of another way to do it.

alfert:
The one-off feature is the entire persistency story of counter examples. With that test case in place, we can also test the regular `counter_strike_test.exs` cases with the same pattern. So, it is then a two-off case, i.e. a main stream feature ;-)

evnu:
That sounds reasonable. Note that I will need some time for the changes, I will be on vacation for a week.

alfert:
Enjoy your vacation!🏖

Von unterwegs  gesendet

evnu:
I finally found some time for this. 157c4ce fixes the message when counter examples are not stored. bdeeda2 verifies that no counter examples are stored if that is requested. 799dace fixes a lint error introduced with #109. Apparently, circleci did not run for that pull request.

alfert:
💚 Like this one!

PR #108, opened 2019-04-16
Removes extra `}` from StateM.DSL examples

adkron:
The examples had a syntax error because there was a `}` hanging
around.


alfert:
Thank you very much 💚

PR #109, opened 2019-04-16
Adds max_size example to StateM.DSL documentation

adkron:
I find that this is a really important aspect when using stateful property-based
Testing. I thought we should document the current path, and have a place to
put something like `more_commands` in later.


alfert:
Valid point. I am currently looking for the root cause, but since I found it and get it fixed, it is really better to have it in the docs.

PR #111, opened 2019-04-29
Propagate :verbose to forall macro

evnu:
This is a follow-up to #48.

When using assertions, :verbose must be set to see the actual output if
an assertion fails. Alas, a user must actually used :verbose twice, once
for the forall macro, and once for a property if forall is used within
that macro. Example:

    property "failing", [:verbose] do
        forall n <- nat(), [:verbose] do
          assert n <= 0
        end
    end

This is currently necessary as we cannot share state between the two
macros. This changeset rectifies that by making the options to the
property available to the forall macro. The approach chosen here is to
use the process dictionary. This is not as functionally clean as one
might want, but it is a simple way to achieve that :verbose is only
needed once in the example above:

    property "failing", [:verbose] do
        forall n <- nat() do
          assert n <= 0
        end
    end

### DIFF
--- a/lib/propcheck.ex
+++ b/lib/propcheck.ex
@@ -383,6 +383,8 @@ defmodule PropCheck do
 
     defp forall_impl(var, rawtype, opts, prop) do
       quote do
+        property_opts = Process.get(:property_opts, [])
+        verbose = :verbose in property_opts || :verbose in unquote(opts)
         :proper.forall(
           unquote(rawtype),
           fn(unquote(var)) ->
@@ -391,7 +393,7 @@ defmodule PropCheck do
             rescue
               e in ExUnit.AssertionError ->
                 stacktrace = System.stacktrace
-                if :verbose in unquote(opts) do
+                if verbose do
                   e |> ExUnit.AssertionError.message() |> IO.write()
                   formatted = Exception.format_stacktrace(stacktrace)
                   IO.puts("stacktrace:\n#{formatted}")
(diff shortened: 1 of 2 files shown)


evnu:
Note that the lint error is fixed in #106.

lib/propcheck.ex
```diff
@@ -383,6 +383,8 @@ defmodule PropCheck do
 
     defp forall_impl(var, rawtype, opts, prop) do
       quote do
+        property_opts = Process.get(:property_opts, [])
```
alfert: That's the first that Joe Armstrong will turn around in his grace. Yuck. Not that I have a better though.

alfert:
I really don't like the process environment, but this seems to be a reasonable solution. At least, I have not better idea. So please do a rebase and after that I will merge this PR. And after that I will make new release.

lib/propcheck.ex
evnu: Yes, I don't like that approach very much either. I spent a lot of time thinking about how to pass the options from a surrounding property into `forall`, but I could not make it work. It is non-obvious to detect within a macro that it is called within another macro and expose the environment of the surrounding macro. At least, the approach here is dead simple and easy to understand.

evnu:
Rebased.

evnu:
Thanks! I think it might be time for a new release, what do you think?

alfert:
> Thanks! I think it might be time for a new release, what do you think?

Yes, this will happen today or tomorrow.

PR #114, opened 2019-08-24
Add a way to set default options

x4lldux:
Removes a lot of boilerplate if all properties in test suite share the same set of options.

Another use case for this is that I'd like to have a quick and quiet tests on my local machine but a big number of verbose tests on CI server. This can be done by passing a function that reads options from system envs.

### DIFF
--- a/lib/properties.ex
+++ b/lib/properties.ex
@@ -81,7 +81,7 @@ defmodule PropCheck.Properties do
   ```
 
   """
-  defmacro property(name, opts \\ [:quiet], var \\ quote(do: _), do: p_block) do
+  defmacro property(name, opts \\ [], var \\ quote(do: _), do: p_block) do
       block = quote do
         unquote(p_block)
       end
@@ -91,6 +91,7 @@ defmodule PropCheck.Properties do
           ExUnit.plural_rule("property", "properties")
           %{module: module} = __ENV__
 
+          module_default_opts = Module.get_attribute(module, :propcheck_default_opts) || [:quiet]
           moduletag = Module.get_attribute(module, :moduletag) |> List.flatten()
           describetag = Module.get_attribute(module, :describetag) |> List.flatten()
           tag = Module.get_attribute(module, :tag) |> List.flatten()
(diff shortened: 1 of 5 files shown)


x4lldux:
Apparently I've missed that compiler shows a strange warning "this clause cannot match because of different types/sizes" when using `property/4` and `default_opts` is a function. I'll push an update with a fix once I figure out what's the cause.

lib/properties.ex
```diff
@@ -106,12 +107,25 @@ defmodule PropCheck.Properties do
           tags = [[failing_prop: tag_property({module, name, []})]]
           prop_name = ExUnit.Case.register_test(__ENV__, :property, name, tags)
           def unquote(prop_name)(unquote(var)) do
+            module_default_opts = unquote(module_default_opts)
```
alfert: I would suggest to put these cases blocks into an ordinary function, which is called here. This eases testing a lot, because you can test the functions independent of the macros. 

Also it is not clear for me why you differentiate between functions and lists. What about maps? They would work as well and are also used for configuring.

alfert: Perhaps this helps also when searching for the root cause for the warning.

x4lldux: Keywords or maps are both static configuration resolved on compilation time, function returns a config during run time.

alfert:
@x4lldux Thank you for this PR, I like this feature. Sorry for being late with my response, but real life requires its time. 

Could you please add a test, which uses this feature? I would at least like to see the one test suite that uses this new feature. You can change one of the existing test suites for this. In the best case you come up with test, that shows the these parameters are really set. This might be a little bit difficult since test that test behaves correctly - lovely meta level :-) Together with a test, I will merge the PR asap.

alfert:
And please do a rebase to put your PR on top of #116, thank you!

x4lldux:
@alfert Done. There aren't actually any tests which check how `property/3` macro processes options so there was nothing for me to reuse. And because of the "meta-levels" there is no direct way to test and so I had to use indirect way by checking process's dictionary.

alfert:
💚 Thanks for this PR! 

To be honest, I am not too happy about the use of the process dictionary which should be avoided. But for now it is more important to have this feature available. Perhaps we find a way to remove the process dictionary.

PR #118, opened 2019-09-27
Allow setting verbose in environment

### LINKED ISSUE #117: Allow setting verbose on the command line
When debugging a property, it is often useful to set the `:verbose` option. Currently, this involves editing the source code. That can be cumbersome,as it involves going back-and-forth between running the tests and setting the flag. What do you think of adding a command line flag, or an environment variable, to override the verbose setting? E.g.

```
mix test /some/property:line --verbose
```

Or, if that is not possible

```
PROPCHECK="--verbose" mix test /some/property:line
```

evnu:
This pull request adds the option to set verbosity using the environment with the `VERBOSE` variable:

```
$ VERBOSE=1 mix test
```

Fix #117

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,11 @@ independent from the build environment, in the file `propcheck.ctex`. The file c
 be removed using `mix propcheck.clean`. Note that this task is only available if PropCheck
 is also available in `:dev`. Otherwise, `MIX_ENV=test mix propcheck.clean` can be used.
 
+### Setting PropCheck into Verbose Mode
+
+Properties in PropCheck can be run in quiet or verbose mode. Usually, quiet is the default. To
+enable verbose mode without requiring to touch the source code, the environment variable `PROPCHECK_VERBOSE`
+can be used. If this is set to 1, the `forall` macro prints verbose output.
 
 ## Links to other documentation
 
--- a/lib/propcheck.ex
+++ b/lib/propcheck.ex
@@ -393,7 +393,9 @@ defmodule PropCheck do
     defp forall_impl(var, rawtype, opts, prop) do
       quote do
         property_opts = Process.get(:property_opts, [])
-        verbose = :verbose in property_opts || :verbose in unquote(opts)
+        env_verbose? = System.get_env("PROPCHECK_VERBOSE") == "1"
+        verbose =
+          :verbose in property_opts || :verbose in unquote(opts) || env_verbose?
         :proper.forall(
           unquote(rawtype),
           fn(unquote(var)) ->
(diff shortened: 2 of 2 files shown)


evnu:
I did not change the README or documentation as I am not sure where to place information on this variable. @alfert If you would like to merge this PR, could you maybe give me a hint were to document this so that others find the option?

alfert:
💚 Thanks, @evnu!

Is `VERBOSE` a good name? It feels very generic. Is something like `PROPCHECK_VERBOSE` more appropriate or is it annoying because you have to type so much?

Documentation is a good point. I suggest to put the docs into `propcheck.ex`: before the `## Acknowledgements` as `## Commandline options` or similar. Additionally, I assume that is a good idea to mention the command line option also in the docs for `forall`. Finally, we should put a small remark in the README under `## Basic Usage and Build Configuration` regarding the configuration options. We should add the command line option at the end, i.e. after discussing the counter examples. 

What do you think? Are these the areas where somebody would look for that information?

alfert:
> Is VERBOSE a good name? It feels very generic. Is something like PROPCHECK_VERBOSE more appropriate or is it annoying because you have to type so much?

Reading your proposal in #117 again, I would suggest to think about the `--verbose` option. Should be similarly easy to access via `System.argv/0`. It feels more natural to append a `--verbose` instead of prepending an environment variable.

evnu:
> What do you think? Are these the areas where somebody would look for that information?

Yes, I also think this should be mentioned in multiple places. Both README and module docs seems appropriate.

> It feels more natural to append a --verbose instead of prepending an environment variable.

I will give that a try. I was wondering if adding such an argument to `mix test` is being a good citizen, as we interfer with the options that `mix test` offers. But I guess this is ok.

evnu:
Hm, actually, `mix test` complains about us not playing fair:

```
mix test test/propcheck_test.exs:79 --verbose
** (Mix) Could not invoke task "test": 1 error found!
--verbose : Unknown option
```

evnu:
> Hm, actually, `mix test` complains about us not playing fair:

So maybe we should go with `PROPCHECK_VERBOSE=1` instead.

evnu:
I renamed `VERBOSE` to `PROPCHECK_VERBOSE` and added remarks to the documentation.

alfert:
👍 go ahead !

Von unterwegs  gesendet

alfert:
@evnu , could you please rebase this PR? Then I will merge and create a new release.

evnu:
> could you please rebase this PR?

Done.

Issue #96, opened 2019-03-18
Fix CI


evnu:
It seems that travis is no longer called when pull requests are submitted. Can this be fixed?

alfert:
I recognized this as well, but I have no clue. My assumption is that this has something to do with the move of travis towards their new services. It is annoying.

evnu:
Apparently, travis still builds successfully on my branches (see [this](https://github.com/alfert/propcheck/pull/94#issuecomment-474110004) for example). Maybe the settings for the repository must be adapted?

evnu:
Maybe [this](https://github.com/travis-ci/travis-ci/issues/9440) issue on travis-ci can provide some hints on what to try.

evnu:
@alfert We could also give circleci a try.

evnu:
In #106:

> Apparently, circleci did not run for that pull request.

@alfert I believe there is some configuration to be done here. It seems that circleci is only building if somebody with a circleci account and PropCheck as a registered project triggers a build. Can you check if you have the following option set for PropCheck:

Settings > Advanced Settings > Build Forked Pull Requests

alfert:
Thanks for the hint. I changed the settings in CircleCI.

PR #120, opened 2019-10-07
Replace circle ci

evnu:
This PR replaces CircleCI again with Travis (#96). CircleCI does not check pull requests, maybe Travis CI does it again.

Note that Travis CI does not store build artifacts (beside caching); we would need to upload crash artifacts to S3.


evnu:
@alfert We originally switched to CircleCI in #96 because Travis did not check pull requests back then. It seems that this has since been fixed; there is a pipeline running for this pull request [here](https://travis-ci.org/alfert/propcheck/builds/594535979?utm_source=github_status&utm_medium=notification).

.travis-scripts/verify_storing_counterexamples.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+#
+# Verify that we can configure storing counterexamples
+#
+
+mix test test/store_counter_examples_test.exs --include will_fail:true && exit 1
+mix test test/verify_counter_examples_test.exs
```
alfert: That's a clever idea to provide a separate build task.

alfert:
@evnu I saw the CI running as well,  good new! Thanks for this PR!

.travis-scripts/verify_storing_counterexamples.sh
evnu: I actually got that from the Travis documentation. :)

Issue #115, opened 2019-09-10
Move counter-examples file to be configured via application configuration (config.exs)


sashaafm:
Hello Alfert, do you have plans or would it be possible to change `prop check` to have the counter-examples files be configure via the standard application configuration (`config.exs` files) instead of via the `Mixfile`?

To us this would be useful to allow running applications depending on `propcheck` on slave nodes via Erlang's `slave` APIs. The problem resides in the slaves receiving the code paths from the code server (Erlang's `code` module), but ignoring the counter-examples file inside the `_build` dir. By having the counter-examples file configured via the application configuration we believe the slave nodes would be able to access the file. We do this with other dependencies who needs files and allow these to be configured via `config.exs`.

Right now, our workaround is to set the `app: false` flag on `mix.exs` dependency declaration for `propcheck`, and manually start `propcheck` in the ExUnit helper file. This allows our property tests to run in the master node, and allows the slave nodes to start up and execute because `prop check` won't look for the counter-examples file.

Thank you for your time and consideration 🙏

evnu:
@sashaafm would you be interested in implementing this after #122 is merged? That way, the new feature can be tested in your slave configuration without too much back and forth.

PR #122, opened 2019-10-14
Enable PROPCHECK_VERBOSE on properties

evnu:
In https://github.com/alfert/propcheck/pull/118, we implemented the environment variable `PROPCHECK_VERBOSE` to set the verbosity on the command line. Alas, this variable does not currently apply to properties. We need to pass the verbosity into the options there to make this work fully.

@alfert This fix feels rather hacky. What do you think, how should this be solved?

### DIFF
--- a/.travis-scripts/verify-verbose.sh
+++ b/.travis-scripts/verify-verbose.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+#
+# Verify that PROPCHECK_VERBOSE works as intended.
+#
+
+search_for="Passed 100 test(s)"
+quiet=$(mix test test/verify_verbose_test.exs)
+verbose=$(PROPCHECK_VERBOSE=1 mix test test/verify_verbose_test.exs)
+
+if echo $quiet | grep "$search_for" -q; then
+    echo >&2 "Found '$search_for' when it should be quiet"
+    echo >&2 "Output:"
+    echo >&2 $quiet
+    exit 1
+fi
+
+if ! echo $verbose | grep "$search_for" -q; then
+    echo >&2 "Did not find '$search_for' when it should be verbose"
+    echo >&2 "Output:"
+    echo >&2 $verbose
+    exit 2
+fi
--- a/.travis-scripts/verify-verbose.sh
+++ b/.travis-scripts/verify-verbose.sh
@@ -3,20 +3,32 @@
 # Verify that PROPCHECK_VERBOSE works as intended.
 #
 
-search_for="Passed 100 test(s)"
-quiet=$(mix test test/verify_verbose_test.exs)
-verbose=$(PROPCHECK_VERBOSE=1 mix test test/verify_verbose_test.exs)
+search_for="OK: Passed 100 test(s)"
+no_global_verbose=$(mix test test/verify_verbose_test.exs)
+global_quiet=$(PROPCHECK_VERBOSE=0 mix test test/verify_verbose_test.exs)
+global_verbose=$(PROPCHECK_VERBOSE=1 mix test test/verify_verbose_test.exs)
 
-if echo $quiet | grep "$search_for" -q; then
-    echo >&2 "Found '$search_for' when it should be quiet"
+count_output_no_global_verbose=$(echo "$no_global_verbose" | grep -c "$search_for")
+count_output_global_quiet=$(echo "$global_quiet" | grep -c "$search_for")
+count_output_global_verbose=$(echo "$global_verbose" | grep -c "$search_for")
+
+if [ "$count_output_no_global_verbose" -ne 1 ]; then
+    echo >&2 "Found '$search_for' more than once when it should exist only once"
     echo >&2 "Output:"
-    echo >&2 $quiet
+    echo >&2 "$no_global_verbose"
     exit 1
 fi
 
-if ! echo $verbose | grep "$search_for" -q; then
-    echo >&2 "Did not find '$search_for' when it should be verbose"
+if [ "$count_output_global_quiet" -ne 0 ]; then
+    echo >&2 "Found '$search_for' when it should be quiet"
     echo >&2 "Output:"
-    echo >&2 $verbose
+    echo >&2 "$global_quiet"
     exit 2
+fi
+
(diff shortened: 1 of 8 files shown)


alfert:
I agree. Spontaneously, I think about a reusable service to determine or retrieve the merged options and about setting the application environment or Config service during application startup. What do you think?

evnu:
> Spontaneously, I think about a reusable service to determine or retrieve the merged options and about setting the application environment or Config service during application startup.

That sounds interesting. Besides the global verbosity, what configurations do you have in mind? If there are no other configuration options which should be held globally for all properties or tests, this might be overkill. In that case, maybe a helper function in the `PropCheck` module (or a `PropCheck.Utils`) would be more suitable?

alfert:
There is also the request for a different storage location for the counter examples. It is a little bit overkill, but not a hack any longer and a proper way of dealing with configuration options. 

Von unterwegs  gesendet

evnu:
To further the discussion, I refactored merging the verbosity options using a `PropCheck.Utils` module in ca8d895 . I was playing around with starting an Agent at start up to hold runtime configuration, but it feels like this does not make too much sense as long as there aren't multiple different configuration options to be considered. In the case for properties and the `forall` macro here, a helper module feels more natural.

evnu:
> There is also the request for a different storage location for the counter examples. It is a little bit overkill, but not a hack any longer and a proper way of dealing with configuration options.

Sorry, I wrote my reply before your comment. Yes, this would be useful for #115. Given the `PropCheck.Utils` module above, I would suggest that we use such a module to perform the actual merging of options, but instead of calling into `System` to retrieve the runtime configuration, we would instead call into a runtime configuration agent. That agent then holds the verbosity option, as well as the location for counter examples, or any future runtime configuration option.

evnu:
I added such an agent in 2d284b2. Let me know if you think that this is the right way forward. Note that the order of processes in the supervisor must be switched (counter examples after this agent) if `CounterStrike` should retrieve options from it.

alfert:
> I added such an agent

Hmm. I thought about using `Application.get_env` which provides all the configuration infrastructure coming from `config.exs`. In our application startup we can add the configuration entry for verbosity from the system's environment.

evnu:
So, you would propose something like this to be called in `app.ex`?

```elixir
defp populate_application_env do
  Application.put_env(:propcheck, :verbose, System.get_env("PROPCHECK_VERBOSE") == 1)
  # .. and add other configuration options here ..
end
```

This seems suitable to me as well, if I understood you correctly.

alfert:
Yes, exactly. For the moment we have only the `PROPCHECK_VEFRBOSE` option to set here. But in the `Utils` module we should also handle the case for #115 which would be available in the app environment if set in `config.exs`. But we can do this in an additional PR as we need to decide how to handle this option.

evnu:
I added `populate_application_env` with 5437f26 and removed the agent. I tested the change locally with this test:

```elixir
defmodule VerboseTest do
  use ExUnit.Case
  use PropCheck

  property "quiet" do
    forall x <- nat() do
      x >= 0
    end
  end
end
```

```
~/tools/propcheck(enable-propcheck-verbose-on-properties ✗) mix test verbose_test.exs 
.

Finished in 0.05 seconds
1 property, 0 failures

Randomized with seed 361789
~/tools/propcheck(enable-propcheck-verbose-on-properties ✗) PROPCHECK_VERBOSE=1 mix test verbose_test.exs 
....................................................................................................
OK: Passed 100 test(s).
.

Finished in 0.07 seconds
1 property, 0 failures

Randomized with seed 208366
```

It might make sense to run this on travis as well. For this, we would need to capture the output twice and compare it. @alfert what do you think, is this worth it? We have precedence with `.travis_scripts/verify_storing_counterexamples.sh` and could solve this similarly.

alfert:
I like this, in particular to ensure that properties and `forall` behave the same under these conditions.

evnu:
> in particular to ensure that properties and forall behave the same under these conditions.

Yes, that is definitely an improvement. Are there other places where this should be done as well that I may have missed?

alfert:
What might be different is the targeted environment, since it uses different generators and overall behaviours. Anything else should be handled by using `forall` and properties.

evnu:
While I am at it: Do you think `PROPCHECK_VERBOSE=0` should disable verbose completely, overriding a `:verbose` setting?

evnu:
I added a test for this feature. I think adding `PROPCHECK_VERBOSE=0` would be nice though to make this symmetric to `PROPCHECK_VERBOSE=1`.

@alfert Do you want the history squashed at some point?

alfert:
Symmetry is nice. Some early mroning thoughts:`PROPCHECK_VERBOSE=0` means it would be globally off by overriding all verbosity settings on the property level. But `PROPCHECK_VERBOSE=1`  only adds `:verbose` to the options and the check in `forall` only asks if `:verbose` exists in the option set.  At least this mechanisms has to change and it is good that we have factored out the `merge_options` function. It will not be enough to add `:quiet` to  the head of list of options.

In short: yes, good idea. And please add it to the docs. Regarding squashing: If you wanna achieve a nice history, then remove the exploratory commits (e.g. for the agent) by interactive rebase. Squashing a feature to a large commit is generally not my favourite.

evnu:
> It will not be enough to add :quiet to the head of list of options.

Right, I did not think of this. I believe this should be handled by enforcing that the first of `:quiet` or `:verbose` wins. Then, prepending either is enough.

evnu:
I added `PROPCHECK_VERBOSE=0` and removed the agent experiment from the history. @alfert can you give this a run?

evnu:
Interesting, the verification script fails for 
0fb5983  on travis, but works locally.

evnu:
> Interesting, the verification script fails for 0fb5983 on travis, but works locally.

So this was due to a different shell. I tried to use `<<<` to pipe variable contents into `cat`. This works here (bash 5.0.011), but not on Travis. I fixed it by using `echo "$output"` instead. Lets see if the pipeline works now.

alfert:
Looks good, thanks @evnu 💚

PR #124, opened 2019-10-14
Fix deprecation warning for Kernel.Typespec.beam_types/1

evnu:
This fixes a deprecation warning for `Kernel.Typespec.beam_types/1`:

```
warning: Kernel.Typespec.beam_types/1 is deprecated. Use Code.Typespec.fetch_types/1 instead
  test/propcheck_test.exs:16
```

As can be seen in [this](https://travis-ci.org/evnu/propcheck/builds/597847058) pipeline, this raises the minimum Elixir version to v1.7.

@alfert This deprecation is not urgent, feel free to close the PR.

### DIFF
--- a/test/propcheck_test.exs
+++ b/test/propcheck_test.exs
@@ -13,7 +13,7 @@ defmodule PropcheckTest do
   @type tagged_stack(t) :: {:stack, [t]}
 
   test "find types in proper_gen.erl" do
-    types = Kernel.Typespec.beam_types(:proper_gen)
+    types = Code.Typespec.fetch_types(:proper_gen)
     refute nil == types
 
     Logger.debug(inspect types, pretty: true)


alfert:
Strange behaviour of Travis: There is no compilation failure or exception, but the counter example storage tests fail.

evnu:
Aren't they expected to fail? The test checks that counterexamples are not stored. Or does https://github.com/alfert/propcheck/blob/master/test/verify_counter_examples_test.exs fail?

alfert:
> Aren't they expected to fail? 

Yes, you are right. I misinterpreted the logs. 

The proper way of handling this would be to differentiate between Elixir < 1.7 and Elixir >= 1.7. The next Proper version will require Elixir 1.6 due to OTP requirements. So we have to live with this glitch for a while.

evnu:
> So we have to live with this glitch for a while.

Yes, that's what I guessed. I think leaving the deprecation warning in until PropCheck requires Elixir >= 1.7 should be ok.

PR #125, opened 2019-10-21
Allow configuring counter examples path in config/

### LINKED ISSUE #115: Move counter-examples file to be configured via application configuration (config.exs)

evnu:
This pull request extends configuration to store counter examples to the application environment. Now, the following is possible:

```elixir
config :propcheck, counter_examples: "location"
```

If the path is configured both in the project definition as well as in the application environment, the application environment takes precedence. As before, it is configured in neither location, it takes a default value.

Fix #115

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,15 @@ def project() do
 end
 ```
 
+Note that the path can also be set as part of the application environment:
+
+```
+config :propcheck, counter_examples: "filename"
+```
+
+If both the project configuration and the application environment are present, the
+application environment is chosen.
+
 Per default, the counter examples file is stored in the build directory (`_build`),
 independent from the build environment, in the file `propcheck.ctex`. The file can
 be removed using `mix propcheck.clean`. Note that this task is only available if PropCheck
--- a/lib/app.ex
+++ b/lib/app.ex
@@ -17,7 +17,7 @@ defmodule PropCheck.App do
         start: {
           PropCheck.CounterStrike,
           :start_link,
-          [Mix.counter_example_file(), [name: PropCheck.CounterStrike]]
+          [[name: PropCheck.CounterStrike]]
         }
       }
     ]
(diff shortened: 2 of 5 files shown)


evnu:
@sashaafm Can you test if this works for you?

evnu:
@alfert `property Default PBT Level 0` [failed again](https://travis-ci.org/alfert/propcheck/jobs/600960159), maybe we should mark it as often failing.. Interestingly, the last times I saw this it happened on Elixir v1.5, but not on the other versions. But this could very well be coincidence.

sashaafm:
@evnu I’m no longer working on the particular project which needed this. I’ll try to find some time to experiment your changes

evnu:
Rebased on current master.

evnu:
Rebased on current master (cleaner history and let travis run again).

evnu:
This has been open for quite some time. I guess we don't actually need this. Closing.

sashaafm:
As indicated in the other issue https://github.com/alfert/propcheck/issues/190, could we get around to merging this?

evnu:
Rebased on master. @sashaafm can you check if this works for you? If so, @alfert would you consider merging this, or should this be handled differently?

alfert:
I restart now the build jobs after merging #191

alfert:
Ah, this cannot work in this way. @evnu, I simply merged via the WebUI the master into this branch. Should only be the github actions.

PR #129, opened 2019-11-12
Translate terms in output from Erlang to Elixir syntax

x4lldux:
PropEr outputs terms in Erlang syntax, and by extension so does
PropCheck. Although it outputs just a subset of Erlang syntax (just the
terms) and reading them is generally not a problem (and even a useful
skill), it can be a pain when trying to copy it to IEx shell. Also it
might discourage new users from using PropCheck.

This commit fixes that. It does this by providing a default function for
`:on_output` option as a translator function. It also touches how
`:verbose` option is processed. That is because PropEr uses _on_output_
functionality for both `:silent` and `:verbose` options by either
supplying a "no print" or "print all in Erlang syntax" functions. That's
why whenever a `:verbose` option is provided to one of functions/macros
we need to substitute `:verbose` for our custom "print all in Elixir
syntax" function - `PropCheck.Utils.elixirfy_output/2`.

### DIFF
--- a/.travis-scripts/verify-verbose-in-elixir-syntax.sh
+++ b/.travis-scripts/verify-verbose-in-elixir-syntax.sh
@@ -9,9 +9,9 @@ seed=0
 elixir_syntax_output=$(mix test --seed 0 test/verify_verbose_elixir_syntax_test.exs)
 
 # linked process crashes
-expected='A linked process died with reason: an exception was raised:
-** (ArithmeticError) bad argument in arithmetic expression'
-if [[ ! $elixir_syntax_output =~ "$expected" ]]; then
+expected1='A linked process died with reason: an exception was raised:'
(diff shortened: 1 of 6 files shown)


x4lldux:
Ehh. It seems that test environment has an older bash version which doesn't support `=~` operator. I'll fix that tomorrow.

alfert:
Thanks a lot for this PR! I never digged deep enough into PropEr to understand/find the implementation of the `:verbose` option, but for sure, the Erlang output was always one the most annoying parts of PropCheck in terms of user-friendliness .

x4lldux:
@alfert Done. 

Few times I forgot to correctly run entire suite of tests, which is also why I'm proposing some changes in #130.

alfert:
💚 Thanks a lot!

PR #130, opened 2019-11-13
RFC: Simplify tests dispatching

x4lldux:
**DO NOT MERGE YET**
Running tests in PropCheck isn't exactly easy. There are few tags tests
that are intended to fail and some which are unstable, additionally
there are few external tests that need be run form a bash shell. So
simply doing `mix test` doesn't work and will just confuse users.

This commit simplifies that by introducing new Mix alias `tests` which
runs both external and regular tests (limited only to reliable tests).

But that also doesn't fix all the issues, because a new developer will
_"ignorantly"_ and out of habit run `mix test` and get themselves in all
sorts of mixed results, not knowing about those `will_fail`, `unstable`
and external tests.

A solution to this would be to alias the `test` Mix command itself or
maybe inverse the test logic and exclude those test by default in
`test_helper.exs` and find a different way to run external tests
alongside Elixir tests... 
This is what I'm proposing to be discussed.


mix.exs
```diff
@@ -72,4 +76,18 @@ defmodule PropCheck.Mixfile do
       {:credo, "~> 1.0.0", only: [:dev, :test], runtime: false}
     ]
   end
+
+  defp test_suite(), do: [
+    &run_external_tests/1,
```
evnu: I would have expected the external tests to run last. Did you have a specific reason to place them first? (I don't have a real preference, just curious)

mix.exs
```diff
@@ -18,6 +18,7 @@ defmodule PropCheck.Mixfile do
      description: description(),
      propcheck: [counter_examples: "_build/propcheck.ctx"],
      aliases: aliases(),
+     preferred_cli_env: [tests: :test],
```
alfert: When was `preferred_cli_env` added to mix? To be honest, I see it here for the first time. Does this require a dependency to a higher Elixir version?

x4lldux: It was added in 1.3.0, PropCheck needs 1.5 so no version change needed.

mix.exs
x4lldux: Just to have coverage report and tests result line (_21 doctests, 55 properties, 17 tests, 1 failure, 23 excluded_) easily visible. I did start with the order you were expecting and then I was scrolling up, so I switched it. The reverse order needs `propcheck.clean` in between though. :confused:

x4lldux:
After digging so more, maybe those external Elixir tests should be tagged `:external` and also excluded from running, or completely moved to a different location - both `*_test.exs` and `*.sh` file. Because it is them that need that additional `propcheck.clean` step, also why run them twice - they were already run by their respective shell scripts.

evnu:
> After digging so more, maybe those external Elixir tests should be tagged :external and also excluded from running, or completely moved to a different location - both *_test.exs and *.sh file. 

That makes sense, we can simply run them manually as part of the corresponding verification shell script.

mix.exs
alfert: TIL: I should read the the beautiful manual more properly

x4lldux:
I've tagged them `:manual` and excluded them by default (alongside unstable and failing tests). Unfortunately running the whole suite still requires doing it via `mix tests` instead of `mix test`. I'm not sure this can be done with out resorting to some hacks.

x4lldux:
I'm closing this - I'm preparing other changes to test that will incorporate this. I don't see any other way for running the whole test suite than creating a special mix alias just for that.
Thank you for input.

Issue #131, opened 2019-11-14
StateM DSL commands with state


x4lldux:
What's the rationale behind `commands/1` generating a tuples with state and a symbolic call? Am I missing something? Why not just a symbolic call like in regular `StateM`?
When PropEr is generating counter-examples, or when printing a "report" with `with_fail/2`  it gets very noisy, especially when state is big.

alfert:
Good question. I need to dig into the code to come with a good answer. This may take some time, since I am quite busy at the moment. Please be patient.

x4lldux:
If that helps, I have a branch with some changes that removes state form the list of generated commands. It looks as it works, but not knowing why it was introduced in the first place, I might have missed something. It introduces a semi breaking change, though: `run_commands` needs an additional argument with module implementing `&initial_state/0` (just like the regular StateM needs). I've added a 1 argument version to keep compatibility, which infers that module from the first generated command - but does not necessarily has to be in the same module as `&initial_sate/0` function.
See https://github.com/x4lldux/propcheck/tree/quieter_statem_dsl

alfert:
I took a short look at the source. I agree that this looks like it is not really needed, since the only situation where it is used (in `execute_commands`), the parameter is ignored. I need to take a look at your implementation though.

x4lldux:
Actually it's used in `run_commands` to get the initial state out of the first command, and also in `is_valid` to do the same. `is_valid` actually calls it a _first_state_ and compares it to _initial_state_ which gets from `mod.initial_state()` -- this comparison is what I don't understand the most, because it looks like a redundant check but often they are there because of some edge cases.

PR #132, opened 2019-11-17
Add pretty reports for StateM

x4lldux:
Conventional way of inspecting the counterexamples generated by stateful
properties involves inspecting the history of commands' return values,
model's state changes and inspecting the property result itself (see
`test/movie_test.exs`). This is not only ugly & hard to read, but also
can be misleading, because depending of the reason of failure the
failing command can be either the last command with the associated
history, or the one before that.

This commit adds a report generator, that prints out a user-friendly and
IEx-shell-friendly (can be safely pasted into the IEx shell) list of
commands, correctly indicating which command was at fault and what kind
of failure was it. Also printing out associated return values and
model's state at each transition.

Compare [inspecting  `&run_commands/2` result](http://sixteencolors.net/paste/x1/c254a987.ans.png) to [new report generator](http://sixteencolors.net/paste/x1/aeb14869.ans.png) -  I've 
introduced 2 bugs: when there are more than 3 users created, asking for
popcorn results in `:no_more_left` and when that happens postcondition
raises an error.

### DIFF
--- a/lib/statem.ex
+++ b/lib/statem.ex
@@ -392,4 +392,23 @@ defmodule PropCheck.StateM do
   """
   defdelegate zip(l1, l2), to: :proper_statem
 
+  @doc """
+  Print pretty report of the failed command run.
+
+  Accepts options:
+  * `return_values` - whether to print return values after each command run
+    (default `true`),
+  * `last_state` - whether section with the last state should be present
+    (default `true`),
+  * `pre_cmd_state` - whether to print state prior to executed command
+    (default `false`),
+  * `post_cmd_state` - whether to print state post executed command
+    (default `true`),
+  * `cmd_args` - whether to print command arguments as literals
+    (default `true`),
+  * `inspect_opts` - options passed to `inspect/2`
+
+  """
+  defdelegate print_report(run_result, cmds, opts \\ []),
+    to: PropCheck.StateM.Reporter
 end
--- a/lib/statem/report.ex
+++ b/lib/statem/report.ex
@@ -1,0 +1,243 @@
+defmodule PropCheck.StateM.Reporter do
+  @moduledoc false
+
+  def print_report({history, state, result}, commands, opts \\ []),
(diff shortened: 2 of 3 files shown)


x4lldux:
The CI fail can be safely ignored, we already discuss it in #130.

alfert:
I like the idea to have a nicer reporting. I thank you already for that effort.

> Compare inspecting &run_commands/2 result to new report generator - 

Can you please provide the correct links (both link to the screen shot)? I would really like to see the differences of the reporting!

x4lldux:
Right, sorry for that. Fixed.

x4lldux:
Also, I was planning to do this for StateM DSL also, but #131 blocks me for now with this.

alfert:
I like this kind of reporting, well done!

I assume that `# -> 4` means that the call return the value `4`? If we allow for `Post state: `, then we should also write the long form `returns: 4` or `return value: 4`. So I don' t need to decipher the output.

x4lldux:
Yes, `-> 4` means the function returned `4`. I thought it universally understood, and easier to spot just by glancing over. I guess it can be changed, but I still like the more concise form better.

alfert:
Could you please rebase it to the current master? Then I would like to merge this PR. Good work, thank you 💚

x4lldux:
@alfert done :)
Though I'd love for it to also work with StateM DSL once #131 is resolved.

alfert:
> Though I'd love for it to also work with StateM DSL once #131 is resolved.

Ok, then let's wait us a little more. I hope that I find some time during the holiday season to invest on #131.

x4lldux:
I've been thinking about it, and now I'm for merging this now. I think it would be better to implement reports for DSL in a separate PR.

alfert:
Please do a rebase, then I will merge the PR.

x4lldux:
@alfert done.

alfert:
Thank you! 💚

PR #134, opened 2019-11-26
Do not crash property if no counter example was returned

### LINKED ISSUE #133: Using `PropCheck.TargetedPBT.exists/2` causes a crash
The `PropCheck.TargetedPBT.exists/2` causes a crash when it returns false.
```Elixir
  property "`exist` crashes on false" do
    exists p <- false do
      p
    end
  end
```
crashes with:
```
  1) property `exist` crashes on false (PropCheck.Test.TargetPathTest)
     test/targeted_path_test.exs:65
     ** (CaseClauseError) no case clause matching: false
     code: property "`exist` crashes on false" do
     stacktrace:
       (propcheck) lib/properties.ex:190: PropCheck.Properties.handle_check_results/5
       test/targeted_path_test.exs:65: (test)
```

evnu:
This fixes the clause error found by @x4lldux when the `exists/2` macro fails.

Fix #133

### DIFF
--- a/lib/properties.ex
+++ b/lib/properties.ex
@@ -164,23 +164,31 @@ defmodule PropCheck.Properties do
 
     proper_opts = PropCheck.Utils.to_proper_opts(opts)
 
-    case CounterStrike.counter_example(name) do
-      :none -> PropCheck.quickcheck(p, [:long_result] ++ proper_opts)
-      :others ->
-        # since the tag is set, we execute everything. You can limit
-        # the amount of checks by using either --stale or --only failing_prop
-        qc(p, proper_opts)
-      {:ok, counter_example} ->
-        # Logger.debug "Found counter example #{inspect counter_example}"
-        result = PropCheck.check(p, counter_example, [:long_result] ++ proper_opts)
-        with true <- result do
+    result =
+      case CounterStrike.counter_example(name) do
+        :none -> PropCheck.quickcheck(p, [:long_result] ++ proper_opts)
+        :others ->
+          # since the tag is set, we execute everything. You can limit
+          # the amount of checks by using either --stale or --only failing_prop
           qc(p, proper_opts)
-        else
-          false -> {:rerun_failed, counter_example}
-          e = {:error, _} -> e
-        end
-    end
-    |> handle_check_results(name, opts, should_fail, store_counter_example?)
+        {:ok, counter_example} ->
+          # Logger.debug "Found counter example #{inspect counter_example}"
+          result = PropCheck.check(p, counter_example, [:long_result] ++ proper_opts)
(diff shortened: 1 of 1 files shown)


alfert:
Credo thinks that you introduced one branch too much to comprehend the function :-) 

Perhaps some refactoring should be advisable (e.g. inside the credo configuration or event the Elixir source code).

evnu:
Refactored and rebased.

lib/properties.ex
```diff
@@ -186,52 +186,92 @@ defmodule PropCheck.Properties do
...
+        |> add_additional_output(opts),
+      expr: nil
+  end
+
+  defp handle_check_results(
+         {:rerun_failed, counter_example},
+         name,
+         opts,
+         _should_fail,
+         _store_counter_example?
```
x4lldux: Maybe should_fail & store_counter_example? should be moved to a map? In most clauses they are not used. Would make code clearer.

evnu: Good idea. I was wondering how to make this a little bit prettier.

evnu: Maybe it would even make sense to put all arguments into a map? We can still match on those parts which are needed? @alfert What do you think?

evnu: I went ahead and turned all arguments into a map with ecc7a08. What do you think, is this better?

x4lldux: I would extract `result` as a separate argument and keep the pipe flow.

evnu: Introduced the pipe again with fe8ed8e.

x4lldux: Looks nice.

evnu: The pipeline failed again in targeted pbt. :man_shrugging:

x4lldux: It's a flaky test, no worries.  I have fix for that in a PR I'll submit later.

alfert: Oh cool, I am very interested into this "take this, flaky test" PR!

Good work, by the way!

alfert:
Do you consider this as done?

lib/properties.ex
x4lldux: @alfert don't be ;) It's an _"engineering"_ solution, not a _"scientific"_ one.

evnu:
> Do you consider this as done?

I think so, but apparently the test coverage for `handle_check_results/2` is not very high when running `mix test --cover --exclude will_fail:true --exclude unstable_test:true`, so I might have missed something during refactoring. So, a review of that is important.

x4lldux:
Looks OK, I didn't see anything missed in it.

alfert:
Then let's merge. Thanks again for your work!

Issue #135, opened 2019-11-28
Allow `property/1` to mark stub tests


evnu:
I sometimes use `test/1` with `ExUnit` to write down a test-to-be-implemented. For example:

```elixir
defmodule Test do
  use ExUnit.Case

  test "I don't know yet what to test here"
end
```

When executing the test, this fails with `not implemented`.

```
  1) test this test is not yet implemented (Test)
     test/test.exs:4
     Not implemented
```

I think this would be a useful addition to the `property` macro as well. Currently, this fails as there is no function `property/1`:

```elixir
defmodule Test do
  use ExUnit.Case
  use PropCheck

  property "this property is not yet implemented"
end
```

Error:

```
== Compilation error in file test/b_test.exs ==
** (CompileError) test/test.exs:5: undefined function property/1
    (stdlib) erl_eval.erl:680: :erl_eval.do_apply/6
    (elixir) lib/kernel/parallel_compiler.ex:237: anonymous fn/4 in Kernel.ParallelCompiler.spawn_workers/7
```

alfert:
I did not know about this feature of `test`, but I like very much and I always wanted to have something similar. So let's provide a similar feature for `property`  together with a nice documentation! 💚

PR #137, opened 2019-12-02
Implement property/1

evnu:
This implements `property/1` as suggested in #135. Example from `properties_test.exs`:

```
~/tools/propcheck(fix-135 ✗) mix test test/properties_test.exs
Compiling 2 files (.ex)


  1) property not yet implemented (PropertiesTest)
     test/properties_test.exs:22
     Not implemented

.
OK: Passed 1 test(s).
..
OK: Passed 1 test(s).
.

Finished in 0.05 seconds
3 properties, 1 failure

Randomized with seed 546517
[1]    7501 exit 1     mix test test/properties_test.exs
```

### DIFF
--- a/lib/properties.ex
+++ b/lib/properties.ex
@@ -127,6 +127,7 @@ defmodule PropCheck.Properties do
       end
   end
 
+  @doc false
   def merge_opts(opts, module_default_opts) do
     module_default_opts = case is_function(module_default_opts) do
                             true -> module_default_opts.()
--- a/lib/properties.ex
+++ b/lib/properties.ex
@@ -286,7 +286,7 @@ defmodule PropCheck.Properties do
   end
 
   @doc """
-  Defines a not implemented property.
+  Defines a not yet implemented property.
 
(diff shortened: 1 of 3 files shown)


lib/properties.ex
```diff
@@ -283,4 +284,25 @@ defmodule PropCheck.Properties do
       {:ok, {_, [{:abstract_code, {_, ac}}]}} = :beam_lib.chunks(beam, [:abstract_code])
       ac |> Enum.map(&:erl_pp.form/1) |> List.flatten |> IO.puts
   end
+
+  @doc """
+  Defines a not implemented property.
```
alfert: Suggestion: `Defines a not yet implemented property.`

evnu: I was wondering about the same thing :) I took it from ExUnit.Case:

>   Defines a not implemented test with a string.

But I will apply your suggestion.

alfert:
I like this one, thanks! 💚

Issue #139, opened 2019-12-13
Filtering list by minimum length fails when used in a property


kerryb:
Firstly, I'm new to both property-based testing and this library, so apologies if this is user error.

I'm trying to write a property for something that's only applicable to lists with at least two elements, but I can't seem to generate values for the test. Here's a reduced example:

```elixir
defmodule MyAppTest do
  use ExUnit.Case
  use PropCheck

  property "for a list with at least two elements, the tail is never empty", [:verbose] do
    forall [_ | tail] <- such_that(l <- non_empty(list(term())), when: length(l) > 1) do
      Enum.any?(tail)
    end
  end
end
```

When I run it, I get a "can't generate" error:

```
my_app$ mix test

Error: Couldn't produce an instance that satisfies all strict constraints after 50 tries.


  1) property for a list with at least two elements, the tail is never empty (MyAppTest)
     test/my_app_test.exs:5
     Property Elixir.MyAppTest.property for a list with at least two elements, the tail is never empty() failed with an error: {:error, :cant_generate}
     code: nil
     stacktrace:
       (propcheck) lib/properties.ex:185: PropCheck.Properties.handle_check_results/4
       test/my_app_test.exs:5: (test)
```

However, when I paste the same code into iex, it seems to work:

```
iex(1)> use PropCheck
PropCheck.TargetedPBT
iex(2)> produce(such_that(l <- non_empty(list(term())), when: length(l) > 1))
{:ok,
 [
   <<58, 117, 253, 188, 1::size(5)>>,
   [-2, {7}, 0, <<190, 21::size(6)>>, 11.336541195778157, []],
   -12,
   {{}, -7.661601316761714, :"", []}
 ]}
```

This is with propcheck 1.2.0, proper 1.3.0, elixir 1.9.4 and erlang 22.1.0, running on macOS 10.15.1.

evnu:
For such cases, I usually build a generator that creates the first two elements and appends a possibly empty list of elements. This has the advantage that it always succeeds, so the constraint will never fail. Example:

```elixir
def at_least_two_integers do
  let [first <- integer(), second <- integer(), many <- list(integer())], do:
    [first, second | many]
end
```

This can then be used in `forall`:

```elixir

property "all integers" do
  forall ints <- at_least_two_integers() do
    Enum.all?(ints, &is_integer/1)
  end
end
```

evnu:
[...]

PR #140, opened 2019-12-16
Clarify `such_that` docs

x4lldux:
Improved docs for situation mentioned in #139

### DIFF
--- a/lib/propcheck.ex
+++ b/lib/propcheck.ex
@@ -718,22 +718,29 @@ defmodule PropCheck do
     This produces a specialization of a generator, encoded as
     a binding of form `x <- type` (as in the let macro).
 
-    The specialization of members of `type` that satisfy the
-    constraint `condition` - that is,
-    those members for which the function `fn(x) -> condition end` returns
-    `true`. If the constraint is very strict - that is, only a small
-    percentage of instances of `type` pass the test - it will take a lot of
-    tries for the instance generation subsystem to randomly produce a valid
-    instance. This will result in slower testing, and testing may even be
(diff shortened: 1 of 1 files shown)


alfert:
Thanks for the new test, that makes it more prominent. 

But, to be honest, I do not really understand what the problem is. Is the approach of Proper that it tries to generate valid data for the small size values (which would not succeed here) and then give up before trying larger size values?

x4lldux:
PropEr generates an instance for the current size. If all constrains pass, *only* then it grows the size and tries the same with a "bigger" instance. In case of #139 it gives up before trying any larger size because it cannot create a valid instance that satisfies the list length constraint.

alfert:
Thanx, now I got it. This means essentially that if the constraint is concerned with the size parameter of the outer generator (here length, which is the size of generator list), then `such_that` is not a good candidate or requires a smart setting of `start_size`.  I propose to add a remark of this kind to the docs and then we are done.

x4lldux:
That doesn't apply just to outer generators nor only to generators that have an easy to grasp "size" property like a list or a binary. For an `integer()` generator small _size_ just means smaller integers being generated and `such_that(n <- int(), when: abs(n) > 100)` would also *"never"* work (if my math is correct there's like a _0.00001_<sup>_constraint_tries_</sup> chance of success for that example).
I'll try to create a better description for that for `such_that` macro.

x4lldux:
@alfert I've pushed the updated docs.

alfert:
💚 Thanks for that PR!

PR #141, opened 2020-01-03
Refactoring of the Statemachine DSL

alfert:
As explained in #131, a refactoring of the stjatemachine DSL removes the state from generated commands. This requires an API change, rendering the `run_commands/1` as deprecated. 

To easier test failing systems, the movie server now has a parameter to introduce failures programmatically. 

Kudos for @x4lldux for the initial steps.

### DIFF
--- a/test/movie_dsl_test.exs
+++ b/test/movie_dsl_test.exs
@@ -6,7 +6,7 @@ defmodule PropCheck.Test.MoviesDSL do
   use PropCheck, default_opts: &PropCheck.TestHelpers.config/0
   use PropCheck.StateM.DSL
   use ExUnit.Case
-  import PropCheck.TestHelpers, except: [config: 0]
+  # import PropCheck.TestHelpers, except: [config: 0]
   require Logger
 
   alias PropCheck.Test.MovieServer
@@ -21,7 +21,7 @@ defmodule PropCheck.Test.MoviesDSL do
     forall cmds <- commands(__MODULE__) do
       trap_exit do
         {:ok, _pid} = MovieServer.start_link()
-        events = run_commands(cmds)
+        events = run_commands(__MODULE__, cmds)
         MovieServer.stop
 
         (events.result == :ok)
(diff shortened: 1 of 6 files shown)


test/support/movie_server.ex
```diff
@@ -59,7 +63,15 @@ defmodule PropCheck.Test.MovieServer do
     :ok
   end
 
-  def handle_call(:popcorn, _from, s), do: {:reply, :bon_appetit, s}
+  defp popcorn_answer(%__MODULE__{popcorn_failure: false}), do: :bon_appetit
+  defp popcorn_answer(%__MODULE__{popcorn_failure: true}), do:
+    Enum.random([:bon_appetit, :bon_appetit, :bon_appetit,
+      :bon_appetit, :bon_appetit, :bon_appetit,
+      :bon_appetit, :bon_appetit, :bon_appetit,
+      :fuck_off])
```
x4lldux: I guess for my _at-home-idea-testing_ phase I could get away with this, but for production code we might be a little family friendly here, and replace this atom ;)

test/movie_dsl_test.exs
```diff
@@ -36,6 +36,27 @@ defmodule PropCheck.Test.MoviesDSL do
       end
     end
   end
+
+  @tag will_fail: true
+  property "server has illegal states" do
```
x4lldux: What's the point of this test? I think it just checks whether the post condition crashes.
If I remember correctly, I introduced the random crashing in `ask_for_popcorn` just as a way to inspect previous implementation.

x4lldux:
With this done, I will start working on DSL version of #132.

test/support/movie_server.ex
alfert: I will see, what the google translator will give me back, so please excuse my french ;-)

test/movie_dsl_test.exs
alfert: Yes, that's true, but this is exactly the reason. An easy toggle of failure, it is automatically excluded by tagging it with `@tag will_fail: true`, but we can check if error messages and the like does behave well.

test/support/movie_server.ex
alfert: Done, it is now satisfying family values.

Issue #142, opened 2020-01-08
StateM.DSL vs PropEr differences


x4lldux:
I've started digging into StateM.DSL while working on generating reports and I've realized it cannot be done, with the same feature set, unless introducing some breaking changes.
But what's more important, translating from tests in _StateM_-ecosystem into _StateM.DSL_-ecosystem was not straight forward because of many differences in interface but also in behaviour:

- when a command crashes StateM.DSL catches it, and continues execution and returns the sequence as successful :confounded: ,
- a crash in postcondition works the same way as a crash in command,
- crash in precondition crashes the test,
- crash in state initialization also crashes the test,
- symbolic datatypes are not fully supported,

While in StateM/PropEr all those crashes just generate a informative return type with a consistent `:exception` tuple. There might be more differences.

So, since we've already introduced some braking changes in #141 and implementing reports would also need to do more of them, why not just reimplement DSL on top of PropEr completely?
This would relieve us from maintaining a replica of the PropEr code and focus on wrapping code instead. 

Of course, that would probably change some (most?!) of the returns types but the DSL would stay mostly the same.

alfert:
Thank you for this review, you seem to be the first, who looked deep enough into this. :green_heart:

> why not just reimplement DSL on top of PropEr completely?

Depends on what you mean exactly. We should fix the crashing, in particular the first one, because it is obviously a bug. Any ideas for supporting symbolic datatypes are highly welcome. But what we cannot do is to implement the DSL on top of PropEr, because PropEr does not have this feature. The DSL is more of a kind of implementing the corresponding QuickCheck feature, since it is a different view on state machines (PropEr and the QuickCheck state machines focus on the states, but the DSL is structured around the transitions/commands). This is the reason for introducing similar and quite complex yet different generators. I would prefer to write a simple wrapper on top of PropEr, but I found no pragmatic way of doing that. At least you would have to implement quite complex macros that generate parts of function bodies that need to be copied to together. But perhaps my thoughts about translating these state machines models into each other were not deep enough and there is some straight and easier way to achieve that.  

If (too many) breaking changes for user code are required to achieve a better reporting including fixing the mentioned inconsistencies and bugs, then I would prefer a second implementation (DSL2) and we deprecate the old one (either as soft or hard deprecation). After a quick look into the Cache example, I do not see at first glance where breaking changes would appear. Can you already foresee where this well be the case?

x4lldux:
> Depends on what you mean exactly.

[...]

PR #143, opened 2020-01-09
Credo + Dialyzer fixes

rodrigues:
Hi @alfert 👋 opening this PR for your appreciation 🙂

- Updates credo, dialyzer and other dev/test deps
- Fixes issues pointed out by credo and dialyzer
- Adds dialyzer to CI (credo already was)

There are still three dialyzer warnings remaining, I'm not sure if these should be fixed or ignored, please tell me what you think about that, or anything else.

Thanks!

```elixir
lib/statem_dsl.ex:476:invalid_contract
The @spec for the function does not match the success typing of the function.

Function:
PropCheck.StateM.DSL.execute_cmd/2

Success typing:
@spec execute_cmd(
  {:set, {:var, _}, {:call, _, _, _}},
  atom() | %{:env => %{{:var, pos_integer()} => _}, :state => _, _ => _}
) ::
  {_, {:call, atom(), atom(), [any()]},
   {{:var, _}, {:exception, _} | {:ok, _} | {:post_condition, _} | {:pre_condition, _}}}

________________________________________________________________________________
test/support/sequential_cache.ex:53:race_condition
The call :ets, insert(PropCheck.Test.Cache, [{1, {_, _}} | {:count, 1, _}, ...]) might have an unintended effect due to a possible race condition caused by its combination with the ets:lookup('Elixir.PropCheck.Test.Cache','count') call in sequential_cache.ex on line 44.
________________________________________________________________________________
test/support/sequential_cache.ex:63:race_condition
The call :ets, insert(PropCheck.Test.Cache, {:count, 0, _}) might have an unintended effect due to a possible race condition caused by its combination with the ets:lookup('Elixir.PropCheck.Test.Cache','count') call in sequential_cache.ex on line 61.
```


x4lldux:
Nice :+1:
Looks like dializer still isn't happy.

rodrigues:
hey @x4lldux , thanks for taking a look!

yes, as I've mentioned in the PR description, I've left still these three dialyzer warnings to see what should be done about them.

about the log result ignoring, so you'd rather match it with `:ok =` then?

x4lldux:
No, it's no different. Using it with every log expression is cumbersome, feels like a boilerplate and is just ugly.
I'd rather  skip `_ =` / `:ok =` altogether, disable `unmatched_return` in test env but leave it in dev env where a human can decide on the matter. Or maybe there's a way for just specifying Logger as an exception in `.dialyzer_ignore.exs` :pray:

rodrigues:
Unmatched returns are annoying with Logger calls, but everywhere else you see one of these warnings, it's usually to want to do something about it, either ignore explicitly or handle the alternate scenarios. It's a bit protocolar, but IMO it's worth the bitter side.

Up to you pals, tell me what you think would be better. Interested to hear also from @alfert on this matter 👍

To avoid bothering about it everywhere, we could also wrap all log calls in a module, handling matching decision (ignore / match), and other potential logging commonalities.

x4lldux:
> Unmatched returns are annoying with Logger calls, but everywhere else you see one of these warnings, you'd often want to do something about it, either ignore explicitly or handle the alternate scenarios.

I'm not saying it's not useful, but rather there's a reason why dialyzer made it disabled by default, and that it might be better to make it enabled in dev env only, and not forcing it in a test env.

rodrigues:
I don't like in my dev flow to have a tool output by default a lot of warnings I shouldn't care about; for me it makes it much harder to use – they end up having so many warnings that it gets hard to spot any new warnings you might want to fix.

So I prefer to aim for the same set of warnings in CI and dev by default, and that both should target 0 warnings output (ignoring / fixing these warnings). A dev can always uncomment ignores to see more, but I'd rather the default in dev would equal CI config.

That's just what I think, hehe, tell me what you think, I'm here just to help you maintain this project, the changes you wish 👍

lib/statem/report.ex
```diff
@@ -192,12 +192,16 @@ defmodule PropCheck.StateM.Reporter do
   defp indent(str, _level, true),
     do: do_indent("# ", str)
 
-  defp indent(str, level, cmt) when is_binary(cmt) and level > length(cmt) do
-    (String.duplicate(" ", level - length(cmt)) <> cmt)
-    |> do_indent(str)
+  defp indent(str, level, cmt) when is_binary(cmt) do
+    length = String.length(cmt)
```
alfert: This seems to handle UTF-8 better

mix.exs
```diff
@@ -69,12 +70,29 @@ defmodule PropCheck.Mixfile do
...
-      {:dialyxir, "~> 0.5.1", only: [:dev, :test]},
-      {:coverex, "~> 1.4.15", only: :test},
-      {:ex_doc, "~>0.16", only: :dev},
-      {:earmark, ">= 0.2.1", only: :dev},
       {:proper, "~> 1.3"},
-      {:credo, "~> 1.0.0", only: [:dev, :test], runtime: false}
+      {:coverex, "~> 1.4", only: :test},
+      {:poison, "~> 3.0", only: :test},
+      {:credo, "~> 1.1", only: [:dev, :test], runtime: false},
+      {:dialyxir, "~> 1.0.0-rc.7", only: [:dev, :test], runtime: false},
```
alfert: When is a 1.0 version to be expected? Any news on that? I would like to stick to a stable version.

alfert:
Hi @rodrigues, thank you for this PR, good work. I have some comments on the source. 

If we go to the new `dialyxir` version, then we are required to upgrade to Elixir 1.6. We need to take that into consideration when merging this PR. 

I have to think about your question regarding the three warning.

mix.exs
rodrigues: Thanks for the feedback, @alfert.

I don't know when they plan to have a `1.0` "stable" release.

But they've been on `1.0.0-rc.x` since 2018, and last version before that (`0.5.1` is from July 2017).

The tool got much better since then, works more nicely with Elixir, I guess they just don't make it `1.0.0` to try resolving some [other issues](https://github.com/jeremyjh/dialyxir/issues) before making it officially "stable".

I'd say rc it's much more stable than 0.5. `rc.7` is actually out since September, me and my team use it in dozens of projects without problems. Last (only) bug I've had with it was when rc.5 got out, and it got resolved really quickly https://github.com/jeremyjh/dialyxir/issues/310

rodrigues:
> If we go to the new `dialyxir` version, then we are required to upgrade to Elixir 1.6. We need to take that into consideration when merging this PR.

I've seen the 1.5 build logs, it complains but it works. As it's just for dev env, I guess we could keep it as is if you wish.

There's a complaint also about a deprecated `--halt-exit-status` option, I'll remove it now 👍

mix.exs
rodrigues: Passing the question along to them as it's indeed a very good point: https://github.com/jeremyjh/dialyxir/issues/379

rodrigues:
I've added [this commit](https://github.com/alfert/propcheck/pull/143/commits/038f633b58037a6acd2c46d5499756c36cf5112c), if you wish, that wraps logging matching convention, making it less cumbersome to use.

lib/logging.ex
```diff
@@ -0,0 +1,14 @@
+defmodule PropCheck.Logging do
```
x4lldux: That's a nice solution.
Not to be picky, but this should be a noun.

alfert:
@rodrigues , can you please rebase the PR and fix the little issues (indentations, Logging -> Logger, ...)?

Regarding dialyzer errors: 
* Do we need a cache for Travis for storing PLTs?
* Can we ignore the race conditions with something like `The call :ets, insert(PropCheck.Test.Cache,.*race condition.*`? 

We stick to Elixir 1.5 and to Dialyxir 1.0-rc*. 

After that I would like to merge the PR (if Travis likes it enough, of course). Good stuff 💚

rodrigues:
thanks @alfert 👍

- rebased
- ignored race conditions for that test cache file
- renamed log module to PropCheck.Logger
- yes, we already have the plts cached (they are kept in `/_build`, already cached in `.travis.yml`)
- please check if indentation is as desired. I made a review and I think now it's ok, it can be better if in the future we put `mix format --check-formatted` in CI and ensure auto-formatting
- 🚧 working now on this last point, the last missing warning from dialyzer:

```
lib/statem_dsl.ex:476:invalid_contract
The @spec for the function does not match the success typing of the function.

Function:
PropCheck.StateM.DSL.execute_cmd/2

Success typing:
@spec execute_cmd(
  {:set, {:var, _}, {:call, _, _, _}},
  atom() | %{:env => %{{:var, pos_integer()} => _}, :state => _, _ => _}
) ::
  {_, {:call, atom(), atom(), [any()]},
   {{:var, _}, {:exception, _} | {:ok, _} | {:post_condition, _} | {:pre_condition, _}}}
```

x4lldux:
@rodrigues With this warning, maybe postpone your work for when #142 is resolved. It might end up as a wasted time if @alfert will agree to my DSL refactor.

test/support/pingpong_master.ex
```diff
@@ -57,10 +57,11 @@ defmodule PropCheck.Test.PingPongMaster do
     #Logger.debug "Player #{inspect name} is waiting round #{counter}"
     receive do
       :ping_pong        -> # Logger.debug "Player #{inspect name} got a request for a ping-pong game"
-          ping(name)
-      {:tennis, from}   -> send(from, :maybe_later)
-      {:football, from} -> send(from, :no_way)
-      msg -> Logger.error "Player #{inspect name} got invalid message #{inspect msg}"
+        ping(name)
+        {:tennis, from}   -> send(from, :maybe_later)
+        {:football, from} -> send(from, :no_way)
```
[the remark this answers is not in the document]
alfert: I will check this one in my editor. It might be something with blanks vs. tabs or so.

alfert:
> ensure auto-formatting

@rodrigues, I won't go for auto-formatting until there is a real case for, which I don't see at the moment. I will take care for this tiny bit later.

lib/statem.ex
```diff
@@ -167,11 +167,11 @@ defmodule PropCheck.StateM do
...
   end
 
-  @type symb_var :: :proper_statem.symb_var
+  @type symb_var :: :proper_statem.symbolic_var()
   @type symbolic_state :: any
   @type dynamic_state :: any
-  @type symb_call :: :proper_statem.symb_call
-  @type command :: :proper_statem.command
+  @type symb_call :: :proper_statem.symbolic_call()
+  @type command :: {:set, symb_var, symb_call} | {:init, symbolic_state}
```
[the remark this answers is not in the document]
rodrigues: `proper_statem:command()` was not exported, that's why we can't use it

test/support/pingpong_master.ex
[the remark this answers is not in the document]
rodrigues: Ah, now I see, thanks. Fix here: https://github.com/alfert/propcheck/pull/145

See, this is one of the things `mix format` would save you from worrying about 😁

PR #144, opened 2020-01-11
Reporter fix

x4lldux:
Small fix. Filename was different than module name.


alfert:
Looks good and prepares for dialyzer. Thanks 💚

PR #145, opened 2020-01-11
Fix indentation

rodrigues:


rodrigues:
I see dialyzer is not passing for 1.6 and 1.7.

That's why build isn't green yet.

I'll handle that in another PR 👍

I should ignore these specific warnings in that file, for all elxiir versions.

alfert:
Perfect, that was the issue I meant.

PR #146, opened 2020-01-11
Ignore error that happens in travis for elixir 1.6 and 1.7

rodrigues:
Their OTP versions are also older (1.6 uses OTP 19), so there may be why we have those weird errors…

Ignoring these warnings for now, we could clean it up in the future when package doesn't support these versions anymore.


alfert:
Please rebase. then I will merge tomorrow.

alfert:
Thanks a lot!

PR #150, opened 2020-01-22
Fix crash in Reporter on passing init state to run_commands

x4lldux:
Passing a third argument to `run_commands/3` in `StateM` or
`StateM.ModelDSL`, an explicit initial state to be used instead of a
callback, and then using `StateM.Reporter` caused a crash on printing it
out.

### DIFF
--- a/lib/statem/reporter.ex
+++ b/lib/statem/reporter.ex
@@ -61,6 +61,8 @@ defmodule PropCheck.StateM.Reporter do
   end
 
   @header String.duplicate("=", 80)
+  defp print_pretty_report(title, cmds_to_print, history, state, [{:init, _} | commands], opts),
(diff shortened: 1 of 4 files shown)


x4lldux:
In both cases fail was due to a flaky test in cache dsl test. Fix for it is in #151

alfert:
Can you please do a rebase for merging this PR?

Thanks a lot!

x4lldux:
Yeah, sure, but there shouldn't be any conflicts with current master.

alfert:
True, but it is more related to have a nice clean history of features in a sequential series. But let's not over-architect that stuff.

alfert:
And again: Thanks for this PR! 💚

PR #151, opened 2020-01-22
Fix a flaky test

x4lldux:
PropEr's command generator may return an empty sequence. We need to
explicitly disallow it here.


alfert:
Oh no, what a silly situation. I am asking me, if we should fix that from the ground up, i.e. apply `non_empty` inside the `command` function. 

Thanks for fixing that!

x4lldux:
I was thinking about that also, but I've decided to not break the current implementation. Running two or three empty command sequences isn't a big problem.

PR #153, opened 2020-03-06
Fix a missing import in ModelDSL

x4lldux:
Fixes an "undefined function <-/2" error when module using ModelDSL does
not import `PropCheck.let/2` macro.


alfert:
Thanks for fixing this bug!

PR #154, opened 2020-03-22
Pretty print StateM command sequence counterexample output

x4lldux:
Instead of printing generated commands counter example as a list of
`{:set, {:var, _}, {:call, _mod, _fun, _args}}` tuples, prints similar
output as `StateM.Reporter` does:

```
  1) property run the misconfigured sequential cache (PropCheck.Test.Cache.DSL)
     test/cache_dsl_test.exs:38
     Property Elixir.PropCheck.Test.Cache.DSL.property run the misconfigured sequential cache() failed. Counter-Example is:
     var6 = PropCheck.Test.Cache.DSL.cache(10, -13)
     var7 = PropCheck.Test.Cache.DSL.cache(1, -17)
     var8 = PropCheck.Test.Cache.DSL.cache(9, -1)
     var9 = PropCheck.Test.Cache.DSL.cache(4, 6)
     var10 = PropCheck.Test.Cache.DSL.cache(-10, 3)
     var13 = PropCheck.Test.Cache.DSL.cache(var_root_key, 12)
     var14 = PropCheck.Test.Cache.DSL.find(10)
     
     
     Counter example stored.
     
     code: nil
     stacktrace:
       (propcheck) lib/properties.ex:226: PropCheck.Properties.handle_check_results/2
       test/cache_dsl_test.exs:38: (test)
```
Unfortunately, even if the failing command is the first, it will print the
entire command list that is being stored as a counter example, and not just the
first one like `StateM.Reporter` would print. But this shouldn't happen often
because the minimization process should reduce the list.

### DIFF
--- a/lib/properties.ex
+++ b/lib/properties.ex
@@ -227,7 +227,7 @@ defmodule PropCheck.Properties do
       message:
         """
         Property #{mfa_to_string(args.name)} failed. Counter-Example is:
-        #{inspect(counter_example, pretty: true)}
+        #{counter_example_inspect(counter_example)}
 
         #{counter_example_message}
         """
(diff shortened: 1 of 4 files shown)


alfert:
Ah, very good. I am working on the parallel stuff and thought about exactly the same thing.

PR #157, opened 2020-03-23
Parallel tests

alfert:
This PR is work in progress and aims at adding the parallel testing capabilities of PropEr in a decent elixirfied version. This includes in particular a good reporting and examples. 

It also employs the instrumentation facilities from #160 for automatic sprinkling of `yield()` into the system under test.


lib/instrument.ex
```diff
@@ -299,64 +299,167 @@ defmodule PropCheck.Instrument do
...
+    Registry: :meta,
+    Registry: :put_meta,
+    Registry: :register,
+    Registry: :select,
+    Registry: :unregister_match,
+    Registry: :unregister,
+    Registry: :update_value,
+  ] ++
+  [ # the name with the dot requires a different encoding
+    {:"Task.Supervisor", :start_link},
```
evnu, line 449: It should be possible to write `"Task.Supervisor": :start_link`.

alfert: Silly me! yes, this is the correct syntax, which I never got right!

PR #160, opened 2020-04-12
Basic instrumentation facilities for parallel testing [WIP]

alfert:
This PR adds instrumentation to PropCheck. It is helpful for parallel testing, since in practice we need to sprinkle `:erlang.yield()` over the place to provoke different runs by the otherwise to deterministic scheduler. 

At the same time, we need such a facility for more complex means, e.g our own scheduler working in the spirit of PULSE for EQC.

### DIFF
--- a/lib/instrument.ex
+++ b/lib/instrument.ex
@@ -1,0 +1,179 @@
+defmodule PropCheck.Instrument do
+  @moduledoc """
+  Provides functions and macros for instrument byte code with additional yields and
+  other constructs to ease testing of concurrent programs and state machines.
+  """
+
+  @doc """
+  Takes the object code of the module, instruments it and update the module
+  in the code server with instrumented byte code.
+  """
+  def instrument_module(mod) when is_atom(mod) do
+
+    # TODO: hwo to convert a beam ast to an Elixir ast?
+    # with :beam_lib we can extract the Erlang AST from the obj_code binary
+    # but how to proceed then?
+    # ast = instrument_functions(mod, obj_code)
+    # {^mod, new_obj_code} = Code.compile_quoted(ast, file)
+    # {:module, _new_mod} = :code.load_binary(mod, filename, new_obj_code)
+    :not_implemented
+  end
+
+  @doc """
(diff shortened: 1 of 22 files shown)


lib/instrument.ex
```diff
@@ -0,0 +1,390 @@
...
+  def encode_value(_unknown), do: throw ArgumentError
+
+  @doc "Encodes a call to `:erlang.yield()"
+  def call_yield do
+    encode_call({:erlang, :yield, []})
+  end
+
+  # Generate the matcher for interestng functions by a macro
+  # This list is taken from the opensource version of Pulse
+  # (https://github.com/prowessproject/pulse-time/blob/master/src/pulse_time_scheduler.erl)
```
evnu, line 304: Do we need to take care of the license here? PropCheck uses GPLv3, Pulse 3-BSD if I am not mistaken.

lib/instrument.ex
```diff
@@ -0,0 +1,390 @@
...
+  def instrumentable_function({:atom, _, :ets}, {:atom, _, :delete_object}), do: true
+  def instrumentable_function({:atom, _, :ets}, {:atom, _, :delete}), do: true
+  def instrumentable_function({:atom, _, :ets}, {:atom, _, :delete_all_objects}), do: true
+  def instrumentable_function({:atom, _, :ets}, {:atom, _, :select_delete}), do: true
+  def instrumentable_function({:atom, _, :ets}, {:atom, _, :match_delete}), do: true
+  def instrumentable_function({:atom, _, :ets}, {:atom, _, :match_object}), do: true
+  def instrumentable_function({:atom, _, :ets}, {:atom, _, :member}), do: true
+  def instrumentable_function({:atom, _, :ets}, {:atom, _, :new}), do: true
+
+  def instrumentable_function({:atom, _, :gen_server}, {:atom, _, :start_link}), do: true
```
evnu, line 319: Note that this does not cover `GenServer`. We might be able to instrument `:gen` though, which is used within Elixir's own `GenServer` implementation.

PR #161, opened 2020-04-13
Employ GitHub Actions as a new CI

alfert:
The results of Travis CI are not shown anymore for whatever reason. So let's try GitHub actions, perhaps they work equally nice.

### DIFF
--- a/.github/workflows/elixir.yml
+++ b/.github/workflows/elixir.yml
@@ -1,0 +1,24 @@
+name: Elixir CI
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Setup elixir
+      uses: actions/setup-elixir@v1
+      with:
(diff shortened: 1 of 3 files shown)


.github/workflows/elixir.yml
```diff
@@ -0,0 +1,30 @@
...
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Setup elixir
+      uses: actions/setup-elixir@v1
```
evnu, line 17: I guess a build matrix to test multiple versions might be useful. See [here](https://github.com/rusterlium/rustler/blob/master/.github/workflows/main.yml#L112) for an example on how to do that for Elixir.

alfert: Yeah, indeed, that would be good. I did struggle with the overall setup, e.g. warnings which are seen as errors in situations where I don't care.  And in particular the dialyzer runs are awfully slow since I also don't use caching of artefacts. So, a helping hand is appreciated!

alfert: I added an issue for that (#163).

PR #164, opened 2020-04-14
Build matrix and Caching

### LINKED ISSUE #163: Make GitHub CI smart
We need a better CI with GitHub Actions to enhance #161: 

* Running on a variety of Elixir versions (i.e. a matrix of Elixir and OTP versions)
* Caching of artefacts for Dialyzer

evnu:
This pull request extends the github actions configuration to

* include a build matrix for multiple versions of Elixir and OTP
* cache dependencies
* cache dialyzer's PLTs in `~/.mix/` and `_build/dev`

Alas, we cannot only cache the PLTs from `_build/dev`, the github cache action requires that a directory is to be cached. We will see if this breaks something.

Fix #163.

### DIFF
--- a/.github/workflows/elixir.yml
+++ b/.github/workflows/elixir.yml
@@ -71,7 +71,7 @@ jobs:
         id: cache-plts
         uses: actions/cache@v1
         with:
-          path: ~/.mix
+          path: _build/dev
           key: ${{ env.ELIXIR_VERSION }}-plt
       - name: Check Credo
         run: mix credo --strict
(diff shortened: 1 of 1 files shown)


evnu:
@alfert There seem to be some dialyzer errors:

```
 lib/instrument.ex:79:unmatched_return
done in 0m2.52s
The expression produces a value of type:

:ok | {:error, _}

but this value is unmatched.

________________________________________________________________________________
lib/instrument.ex:80:unmatched_return
The expression produces a value of type:

:ok | {:error, _}

but this value is unmatched.
```

alfert:
Oh wow, this solution looks very different from mine. GitHub Actions seems to be a more complex universe than what I learned in the last couple of days. Thanks for that!

I will take care for the Dialyzer Bug in `instrument.ex`.

PR #165, opened 2020-05-14
Allow referring to variables declared in generator expressions using pin operator.

Ecialo:
Greetings!

Recently I've started using PropCheck in my small project and quickly ran into a clumsy creating generators using the `let` macro. Every time I have to use something drawn from a generator as parameter in another generator, I was forced to declare another `let` block. But I've found some courage, solved this problem and now I'm sending this PR. 

Now you can refer to declared vars with `^`. It looks like this:

```elixir
nondec =
  let [m <- integer(^l, :inf), l <- integer(), h <- integer(^m, :inf)] do
    {l, m, h}
  end
```
I have also updated docs and wrote some tests for this feature. Credo sees no problems with this code.

Allow referring to variables declared in generator expressions using pin operator.


alfert:
Cool, I like this idea! I will take a look at the code over the weekend.

lib/propcheck.ex
```diff
@@ -678,18 +692,28 @@ defmodule PropCheck do
     """
     defmacro let({:<-, _, [var, rawtype]}, generator) do
         [{:do, gen}] = generator
-        quote do
-            :proper_types.bind(unquote(rawtype),
-                fn(unquote(var)) -> unquote(gen) end, false)
-        end
+        prop_bind(var, rawtype, gen)
```
alfert, line 695: Good point to refactor that out

alfert:
My major concern with this PR is that you added your own specialised graph data structures and algorithms for analysing dependencies. How about using a dedicated library for that? I would recommend `Graph` over `:digraph`, which I will incorporate for analysing the concurrency stuff in the near future. 

Otherwise it looks great. 💚

Ecialo:
Thanks for advice! I've moved to libgraph and it definitely helped to discard lot of code.

lib/propcheck.ex
```diff
@@ -663,6 +663,20 @@ defmodule PropCheck do
         ...>   end)
         true
 
+    You also can refer variables declared in var scope with `^` until they
+    not form cycle or mix with var with same name from outer scope.
```
alfert, line 667: I would go for this: 

You als can refer *to* variables declared in *the current `let`* scope with `^`. Such variables are not allowed to form a dependency cycle . You cannot apply `^` to variables declared outside of the `let` macro.

lib/propcheck.ex
```diff
@@ -663,6 +663,20 @@ defmodule PropCheck do
...
+
+        iex> use PropCheck
+        iex> nondec =
```
alfert, line 670: `nondec` == `non_decreasing`? If yes, then change it please to make it easier to grasp your example.

lib/propcheck/utils.ex
```diff
@@ -142,4 +142,60 @@ defmodule PropCheck.Utils do
   end
   defp parse_io_spec_format(char_or_spec), do: char_or_spec
 
+  def toplevels(graph) do
+    with(
```
alfert, line 146: Are the brackets after the `with` required? Looks strange, but may be correct.

lib/propcheck/utils.ex
```diff
@@ -142,4 +142,60 @@ defmodule PropCheck.Utils do
...
+  end
+
+  def find_all_vars(block) do
+    block
+    |> Macro.postwalk([], &finder/2)
+    |> elem(1)
+    |> Enum.uniq()
+  end
+
+  defp finder(block, acc) do
```
alfert, line 177: what does `finder` find? I would prefer `var_finder`.

lib/propcheck/utils.ex
```diff
@@ -142,4 +142,60 @@ defmodule PropCheck.Utils do
...
+          _ -> {block, [pinned | acc]}
+        end
+      {var, _, args} when not is_list(args) ->
+        {block, [var | acc]}
+      _ -> {block, acc}
+    end
+  end
+
+  def unpin_vars(block) do
+    replacer = fn b ->
```
alfert, line 192: what about `var_replacer` instead of `replacer`?

test/let_test.exs
```diff
@@ -183,4 +183,58 @@ defmodule PropCheck.Test.LetAndShrinks do
 
   end
 
+  defmodule LetChainingTest do
+    use ExUnit.Case
+    use PropCheck
+
+    property "seq of 3 nondecreasing integers in one let" do
+      nondec =
```
alfert, line 191: I would again change towards `non_decreasing` instead of `nondec`

test/let_test.exs
```diff
@@ -183,4 +183,58 @@ defmodule PropCheck.Test.LetAndShrinks do
...
+          {l, m, h}
+        end
+      forall {l, m, h} <- nondec do
+        l <= m and m <= h
+      end
+    end
+
+    property "let with functions inside" do
+      p2 = fn x -> x + 2 end
+      nondec = let [
```
alfert, line 202: see above

test/let_test.exs
```diff
@@ -183,4 +183,58 @@ defmodule PropCheck.Test.LetAndShrinks do
...
+      end
+      forall {a, b, c} <- nondec do
+        a <= b and b <= c
+      end
+    end
+
+    property "mix of inner and external vars in one let" do
+      a = 1
+      b = 10
+      nondec = let [
```
alfert, line 217: see above

alfert:
Thanks for the move to `:libgraph`. I have some remarks attached the diffs. 

> But I have no idea how to resolve this conflicts in mix.lock :о(

If I see that right, then the only significant change is the reference to `libgraph`.  But you are not last version of the master branch. Please rebase by the following: 

  * change to your master branch with `git checkout master`
  * do pull with `git pull --rebase`  
  * change to your branch with `git checkout let_chaining`
  * rebase with `git rebase master`

Now, two things may happen: The conflict is gone and everything is fine. 
Or: there is still a conflict. In this case choose your version (with libgraph). After that do a `mix get.update --all` to have a fresh update. Now there should be no conflicts.  
 
Good work, thanks!

mix.exs
```diff
@@ -71,6 +71,7 @@ defmodule PropCheck.Mixfile do
   defp deps do
     [
       {:proper, "~> 1.3"},
+      {:libgraph, "~> 0.7"},
```
alfert, line 74: The latest version of `:libgraph` is `0.13.13`. Please upgrade!

lib/propcheck.ex
Ecialo: Thank you >_>

lib/propcheck/utils.ex
Ecialo: Yes, they required for such formatting.

Ecialo:
"The pattern can never match the type." problem mostly because of libgraph type spec mismatch. `Graph.topsort/1` specified as `[vertex()]`, but actually it is `[vertex()] | false`.

alfert:
> "The pattern can never match the type." problem mostly because of libgraph type spec mismatch. Graph.topsort/1 specified as `[vertex()]`, but actually it is `[vertex()] | false`.

Hmm, might be, but dialyzer complaints about something way more simple: 

    lib/propcheck.ex:712:pattern_match 
    The pattern can never match the type. 
    Pattern: {:error, _msg}
    Type:  {:ok, [atom()]}

Just remove line 712 in `propcheck.ex` where you match for `{:error, _msg}` and dialyzer should be happy.

alfert:
By the way: you did not do a `git rebase` but a `git merge`, this includes all newer  commits from the master branch into your branch, which is not intended. With rebase, your branch would be move towards the top-level element of master, i.e. as if you started your branch after the latest commit of master. This is an extremely helpful because only your commits are in your feature branch, so it remains clean and the history is linear: There is at most one branch parallel to master . 

If you are confident (or adventurous :-) ) enough with git, then you can reorder the commits with interactive rebase and then remove the merge. After that you would a `git rebase `as explained above. What I missed to explain is that after a `git rebase` on already published branch (i.e. on a PR), you need to force the `git push`. Usually a `git push --force-with-lease` is enough (unless you are working with several people on the same branch). If you are in doubt, then leave it as it is.

Ecialo:
> By the way: you did not do a git rebase but a git merge, this includes all newer commits from the master branch into your branch, which is not intended.

I did `git rebase`, then push without `--force`, then I manually merge mix.lock with master branch in github web ui (it can be noticed this by a small green check mark nearby that commit). I definitely didn't expect 40 new commits :оD

> Just remove line 712 in `propcheck.ex` where you match for `{:error, _msg}` and dialyzer should be happy.

Ok, but it lead to `CaseClauseError` instead of something more meaningful. I better hide this line behind the `#`.

> If you are confident (or adventurous :-) ) enough with git, then you can reorder the commits with interactive rebase and then remove the merge.

I'm not so confident for this :о(, but I'll make one more attempt.

alfert:
@Ecialo Is there an activity pending from your side? Otherwise I would merge the PR in the next hours.

Ecialo:
> @Ecialo Is there an activity pending from your side? Otherwise I would merge the PR in the next hours.

Nope. In the end, I lost to git in an unequal fight >_>

Thank you for awesome tool and your assist with this PR :о)

alfert:
Thanks for this cool new feature! 💚

PR #167, opened 2020-05-16
Let-Chaining with pinned variables

alfert:
This is a replay of PR #165 which had an unlucky merge operation distorting the history. 

All kudos for this code goes to @Ecialo !


alfert:
I cherry-picked all commits from @Ecialo and then reseted the master branch back to the former PR. After that, I could push the repair-branch and start this PR. After the checks are all green, I can merge this PR and we have again a simple, easy history including this very cool feature.

PR #170, opened 2020-07-22
Make the changelog links valid on the respective pages

oo6:


alfert:
Thanks for identifying the glitch and for this PR. 
Could you please use relative paths? There is no need to link to my repository (in case of a fork) or to hex docs (when using e.g. Dash)

PR #176, opened 2020-10-13
Elixir 1.11: :iex should be listed in :extra_applications

ahamez:
The warning was:
```
warning: IEx.Config.color/1 defined in application :iex is used by the current application but the current application does not directly depend on :iex. To fix this, you must do one of:

  1. If :iex is part of Erlang/Elixir, you must include it under :extra_applications inside "def application" in your mix.exs

  2. If :iex is a dependency, make sure it is listed under "def deps" in your mix.exs

  3. In case you don't want to add a requirement to :iex, you may optionally skip this warning by adding [xref: [exclude: IEx.Config] to your "def project" in mix.exs

  lib/statem/reporter.ex:385: PropCheck.StateM.Reporter.syntax_colors/0
```

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -44,7 +44,8 @@ defmodule PropCheck.Mixfile do
   # Type `mix help compile.app` for more information
   def application do
     [applications: [:logger, :proper, :libgraph],
-     mod: {PropCheck.App, []}]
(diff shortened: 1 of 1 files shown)


alfert:
Thanks for identifying and solving this flaw. 💚

PR #177, opened 2020-10-14
Use __STACKTRACE__ macro rather than the deprecated System.stacktrace()

ahamez:
Starting with Elixir 1.11, using procheck generates many warnings:
```
warning: System.stacktrace/0 is deprecated, use __STACKTRACE__ instead
```

Requires Elixir 1.7 as minimal version.

### DIFF
--- a/lib/propcheck.ex
+++ b/lib/propcheck.ex
@@ -429,20 +429,17 @@ defmodule PropCheck do
               unquote(prop)
             rescue
               e in ExUnit.AssertionError ->
-                stacktrace = System.stacktrace
                 if verbose? do
                   e |> ExUnit.AssertionError.message() |> IO.write()
-                  formatted = Exception.format_stacktrace(stacktrace)
+                  formatted = Exception.format_stacktrace(__STACKTRACE__)
                   IO.puts("stacktrace:\n#{formatted}")
                 end
                 false
               e ->
-                stacktrace = System.stacktrace
-
                 if detect_exceptions? do
                   output = """
                   PropCheck detected an error:
-                  #{Exception.format(:error, e, stacktrace)}
+                  #{Exception.format(:error, e, __STACKTRACE__)}
                   """
 
                   if output_agent != nil do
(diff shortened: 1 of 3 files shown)


lib/propcheck.ex
```diff
@@ -429,20 +429,17 @@ defmodule PropCheck do
               unquote(prop)
             rescue
               e in ExUnit.AssertionError ->
-                stacktrace = System.stacktrace
                 if verbose? do
                   e |> ExUnit.AssertionError.message() |> IO.write()
-                  formatted = Exception.format_stacktrace(stacktrace)
+                  formatted = Exception.format_stacktrace(__STACKTRACE__)
```
alfert, line 434: I would suggest a downward compatible solution: Check the version of Elixir at compile time in a macro and put it into here. Something like 

```` elixir
    # outside a function: 
    if Version.match(System.version, ">= 1.7.0") do 
        @STACK __STACKTRACE__
    else 
        @STACK System.stacktrace()
    end
````

This can be used instead of `__STACKTRACE__`.

**Comment by alfert** [comment]
@ahamez For the rebase, please take a look at the README explaining that stuff.

**Comment by ahamez** [comment]
OK, I was wondering if the compatibility with Elixir 1.5 was just a leftover. I'll give the macro a try in a week or two!

**Comment by alfert** [comment]
@ahamez, if everything we change is the compatibility for Elixir 1.11, then I doubt that it is worth a full minor version. The API effectively does not change. 

Looking forward to your solution!

File: lib/propcheck.ex (hunk 1 above)
Comment (flowerett) [reply lib/propcheck.ex:434]: Seems it's not possible to use that macro outside the `try-catch` clause. [(elixir-lang)](https://github.com/elixir-lang/elixir/blob/master/lib/elixir/lib/kernel/special_forms.ex#L718)

Defining inside the function like it was before:
```
rescue
  e in ExUnit.AssertionError ->
    stacktrace = if Version.match?(System.version, ">= 1.7.0") do
      __STACKTRACE__
    else
      System.stacktrace
    end

    if verbose? do
      e |> ExUnit.AssertionError.message() |> IO.write()
      formatted = Exception.format_stacktrace(stacktrace)
      IO.puts("stacktrace:\n#{formatted}")
    end
...
```

will still raise warnings, e.g.
```
warning: System.stacktrace/0 is deprecated, use __STACKTRACE__ instead
  test/store_counter_examples_test.exs:14
```

alfert: Yes, I got the same results with my approach. That your approach does not work is clear, since both branches of the `if` are compiled hence the warning - my approach was trying to be something like conditional compiling. That would work, if `__STACKTRACE__` is allowed outside of a try-catch clause. So let us look for a different approach.

ahamez:
@alfert It seems we can close this PR as https://github.com/alfert/propcheck/pull/178 is a superset. Is it OK with you?

alfert:
@ahamez yes, absolutely. Thank you for your initiative! 💚

PR #178, opened 2020-10-16
Fix deprecations warnings in Elixir 1.11.x

flowerett:
Hello,
This is a fix for all deprecation warnings that PropCheck raises in Elixir 1.11

It's based on this PR: https://github.com/alfert/propcheck/pull/177. I was trying to help @ahamez to fix the issue but ended up in a major refactoring since there are other warnings raised in 1.11

What has been done and some rationale about it:
- [x] 1. drop support for Elixir < 1.7
Unfortunately, I failed to find a way to keep both `__STACKTRACE__` and `System.stacktrace()`
Seems like starting from OTP/23 using `System.stacktrace()` will always raise a warning which is pretty annoying.
I see that some libraries already support only Elixir >= 1.7, 

- [x] 2. In Elixir 1.11 credo failed with a weird error: [travis build](https://travis-ci.org/github/flowerett/propcheck/jobs/736402833)
The easiest way to solve it appeared to update the `credo` itself.

- [x] 3. Support of OTP/23 is not possible without updating the `:proper` library (stack traces will not be visible).
The fix is already merged for `:proper` https://github.com/proper-testing/proper/pull/226/files but updating PropCheck will need an additional effort and will be done in a separate PR

- [x] 4. Github workflows refactored and updated to use the latest Elixir/OTP
We agree to get rid of Travis completely as it is not maintained anymore. 
A left one lint stage that uses Elixir 1.9 and OTP 22.3 and added a new matrix of 5 Elixir/OTP versions. Each stage will use its own cache for dependencies.

- [x] 5. Some methods got deleted completely as they were not used anywhere.


mix.exs
```diff
@@ -5,7 +5,7 @@ defmodule PropCheck.Mixfile do
   def project do
     [app: :propcheck,
      version: "1.2.3-dev",
-     elixir: "~> 1.5",
+     elixir: ">= 1.7.0",
```
alfert, line 8: This would allow Elixir 2.x. We should not be so optimistic :-) Please change to `~> 1.7`, which requires a 1.x version of Elixir.

alfert:
Wow, this is quite a mouthful! Thank you for your work! 

Regarding Travis: This is an old and no longer maintained CI platform, since we moved to CircleCI and nowadays to GitHub Actions. Please remove the `.travis.yml`. 

It seems that we are required to move to Elixir 1.7. This I will release a new minor version such that older Elixir system can still rely on the minor version and can get bugfixes, if required.

alfert:
Regarding 3.: 

> The failing test is: test/verify_verbose_elixir_syntax_test.exs:38

Did you try the test locally on your machine, i.e. by running `PROPCHECK_VERBOSE=1 mix test --include manual --include will_fail test/verify_verbose_elixir_syntax_test.exs --seed 0` ? Does this work and creates the proper output?

flowerett:
> Did you try the test locally on your machine, i.e. by running `PROPCHECK_VERBOSE=1 mix test --include manual --include will_fail test/verify_verbose_elixir_syntax_test.exs --seed 0` ? Does this work and creates the proper output?

Yes, this can be reproduced locally.

elixir          1.11.1
erlang       22.3.4
```
mix test_ext
...
PROPCHECK_DETECT_EXCEPTIONS:  	---> OK
PROPCHECK_DETECT_EXCEPTIONS: 0 	---> OK
PROPCHECK_DETECT_EXCEPTIONS:  	---> OK
PROPCHECK_DETECT_EXCEPTIONS: 1 	---> OK
```

elixir          1.11.1
erlang          23.1.1
```
PROPCHECK_DETECT_EXCEPTIONS:  	---> OK
PROPCHECK_DETECT_EXCEPTIONS: 0 	---> OK
PROPCHECK_DETECT_EXCEPTIONS:  	---> OK
PROPCHECK_DETECT_EXCEPTIONS: 1 	---> OK
Raised exception with stacktrace not found
```
So the last test fails, but with the wrong output (without a stacktrace)

What is even more important - I can reproduce it in the current master. So this is probably an incompatibility with OTP/23 which is not related to the changes I did.

lib/propcheck.ex
```diff
@@ -429,20 +429,17 @@ defmodule PropCheck do
               unquote(prop)
             rescue
               e in ExUnit.AssertionError ->
-                stacktrace = System.stacktrace
```
alfert, line 432: Perhaps it is safer to call here `__STACKTRACE__` instead of a few lines later such that the stack properly defined by the caller of the exception?

alfert:
> But not sure how can it be invoked, I checked all related Elixir and proper files, seems the function was deprecated since OTP/21 and is already properly handled everywhere.

See me inline-comment: Could it be that the stack trace is required to be fetch as first action inside the `catch` block? Or is an error within Elixir? I do not find something similar to the extended `catch` patten of Erlang for Elixir in the docs.

lib/propcheck.ex
```diff
@@ -429,20 +429,17 @@ defmodule PropCheck do
...
                 if verbose? do
                   e |> ExUnit.AssertionError.message() |> IO.write()
-                  formatted = Exception.format_stacktrace(stacktrace)
+                  formatted = Exception.format_stacktrace(__STACKTRACE__)
                   IO.puts("stacktrace:\n#{formatted}")
                 end
                 false
               e ->
```
flowerett, line 438: To be more precise, the test is failing here.

lib/propcheck.ex
```diff
@@ -452,15 +449,13 @@ defmodule PropCheck do
                   end
                 end
 
-                reraise e, stacktrace
+                reraise e, __STACKTRACE__
```
flowerett, line 452: And then, an exception got reraised here. I even tried to remove the whole try-catch clause - the same result.
This makes me think that the issue is not in the PropCheck library or module, but most likely in the erlang's :proper.forall/2

flowerett: I wonder if this macro defined properly:
9e423393d8295b2faa6e1b384b71e/include/proper_internal.hrl#L49

flowerett:
@alfert with help of @adriankumpf I discovered that the issue with tests is indeed in the `:proper` library.
Actually, it's way outdated and version 1.3.0 was released 25 Jul 2018 and it doesn't support the latest OTP.
There is also a fix for the problem, merged to the master branch:
https://github.com/proper-testing/proper/pull/226/files#diff-854de97b58cf5f042d4bb54a95e826c5687ee8030dda92363bd99dca8d30d33bL49

I've tested PropCheck with the master of `:proper` and that particular test pass. This is good news.

But there are also bad 😄 . The API of `:proper` changed, some functions got renamed, some even deleted. A lot of other tests failed. I've started the work for updating the `:proper` as well and already fixed some tests, but not sure if we want to mix it with this PR.
The branch: https://github.com/flowerett/propcheck/tree/update-proper
Last commit: https://github.com/flowerett/propcheck/commit/84007c41629aff5c7875fb8a87e16decc5e9c0ea

My suggestion would be to create a separate issue and PR for updating the `:proper`.
As to this PR, I can suggest skipping tests for OTP/23 and merge it as is (without releasing a new version).
What do you think?

alfert:
Thanks for your effort in solving this puzzle! 

I agree, that we should not mix these two topics up. So please fix the current PR if required, then we can merge it properly. 

If I understand the situation correctly, then we have with PR the opportunity of using Elixir 1.11.x without deprecations warnings. But in order to see stack traces, we need to stick to OTP 22 due to a bug in `:proper`. I assume this is worth a release or do you expect that those running Elixir 1.11.x also use the most up-to-date OTP?

flowerett:
> I assume this is worth a release or do you expect that those running Elixir 1.11.x also use the most up-to-date OTP?
Well, we were one of those who updated to 1.11 and OTP/23 at the same time. But of course, I can't speak for everyone.

So I don't mind releasing the new version, but it's up to you.
It can be just mentioned in the release notes that OTP/23 is not supported yet and an update is in progress.

At the same time, I'll try to create an issue and to bind it to the new PR.

flowerett:
@alfert I've updated PR, it is ready for review and the pipeline now is green.
Thank you!

alfert:
@flowerett Thanks a lot for your work! 💚

PR #179, opened 2020-10-25
Update proper library, Fix OTP/23 issue

flowerett:
This is a WIP for updating the `:proper` to the latest version (from master as it was not released since 2018)

You can find more information on the issue page: https://github.com/alfert/propcheck/issues/180

TODO:
- [x] lint pass
- [x] test pass for all OTP versions
- [x] issue in PropEr fixed: [Pull Request](https://github.com/proper-testing/proper/pull/254)
- [x] test_ext pass for all OTP versions

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -78,7 +78,7 @@ defmodule PropCheck.Mixfile do
   # Type `mix help deps` for more examples and options
   defp deps do
     [
-      {:proper, "~> 1.3"},
+      {:proper, "~> 1.3", github: "proper-testing/proper"},
       {:libgraph, "~> 0.13"},
       {:coverex, "~> 1.4", only: :test},
       {:credo, "~> 1.4", only: [:dev, :test], runtime: false},
--- a/test/verify-verbose-in-elixir-syntax.sh
+++ b/test/verify-verbose-in-elixir-syntax.sh
@@ -11,6 +11,7 @@ elixir_syntax_output=$(PROPCHECK_VERBOSE=1 mix test --include manual --include w
 # linked process crashes
 expected1='A linked process died with reason: an exception was raised:'
 expected2='** (ArithmeticError) bad argument in arithmetic expression'
+# pass
 if ! echo "$elixir_syntax_output" | grep -FA1 "$expected1" | grep -qF "$expected2"; then
     echo >&2 "Crash report from linked process not found"
     echo >&2 "Output:"
@@ -20,6 +21,7 @@ fi
 
 # linked process kills it self
 expected='A linked process died with reason :killed.'
+# pass
 if ! echo "$elixir_syntax_output" | grep -qF "$expected"; then
     echo >&2 "Crash report from linked process not found"
     echo >&2 "Output:"
@@ -28,7 +30,11 @@ if ! echo "$elixir_syntax_output" | grep -qF "$expected"; then
 fi
 
 # collect prints Elixir syntax
-expected='100% %{test: VerifyVerboseElixirSyntaxTest}'
+# TODO fix & uncomment:
+# fails returned result - '*****% %{test: VerifyVerboseElixirSyntaxTest}'
+# previous result:
+# expected='100% %{test: VerifyVerboseElixirSyntaxTest}'
+expected='*****% %{test: VerifyVerboseElixirSyntaxTest}'
 if ! echo "$elixir_syntax_output" | grep -qF "$expected"; then
     echo >&2 "Collected categories not found"
     echo >&2 "Output:"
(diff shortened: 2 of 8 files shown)


test/verify-verbose-in-elixir-syntax.sh
```diff
@@ -28,7 +30,11 @@ if ! echo "$elixir_syntax_output" | grep -qF "$expected"; then
 fi
 
 # collect prints Elixir syntax
-expected='100% %{test: VerifyVerboseElixirSyntaxTest}'
+# TODO fix & uncomment:
+# fails returned result - '*****% %{test: VerifyVerboseElixirSyntaxTest}'
+# previous result:
+# expected='100% %{test: VerifyVerboseElixirSyntaxTest}'
+expected='*****% %{test: VerifyVerboseElixirSyntaxTest}'
```
flowerett, line 37: so far found only this change, but should be also working:
https://github.com/proper-testing/proper/commit/8b600da8987d9015809c2b9e59db6998e3008df8

it's not clear why '*****' is printed instead of digits

flowerett: but sometimes it's working:
```erlang
io:format("~5.2f\%~n", [100 * 20 / 27]).
74.07%
```

flowerett: and it totally makes sense, as erlang returns:
```erlang
100 * 10 / 10.
100.0
```
so we need:
```erlang
39> io:format("~5.1f\%~n", [100 * 10 / 10]).
100.0%
```
otherwise io:format will return wrong result:
```erlang
io:format("~5.2f\%~n", [100 * 10 / 10]).
*****%
```

flowerett: Should be ?
```erlang
67> io:format("~.2f\%~n", [100 * 1 / 30]).
3.33%
ok
68> io:format("~.2f\%~n", [100 * 10 / 30]).
33.33%
ok
69> io:format("~.2f\%~n", [100 * 20 / 30]).
66.67%
ok
70> io:format("~.2f\%~n", [100 * 30 / 30]).
100.00%
```

flowerett: So for success tests, it fails as `100.00` has 6 symbols. 🤦 

Check is pretty simple, fix is also pretty obvious:
```erlang
io:fwrite("printing with different precisions~n"),
io:fwrite("====== 5.2 can fail =====~n"),
io:fwrite("~5.2f ~n", [1000 * 30 / 70]),
io:fwrite("~5.2f ~n", [100 * 30 / 30]),
io:fwrite("~5.2f ~n", [100 * 18 / 30]),
io:fwrite("~5.2f ~n", [100 * 1 / 30]),
io:fwrite("===== 6.2 ok =====~n"),
io:fwrite("~6.2f ~n", [1000 * 30 / 70]),
io:fwrite("~6.2f ~n", [100 * 30 / 30]),
io:fwrite("~6.2f ~n", [100 * 18 / 30]),
io:fwrite("~6.2f ~n", [100 * 1 / 30]),
io:fwrite("===== 5.1 ok =====~n"),
io:fwrite("~5.1f ~n", [1000 * 30 / 70]),
io:fwrite("~5.1f ~n", [100 * 30 / 30]),
io:fwrite("~5.1f ~n", [100 * 18 / 30]),
io:fwrite("~5.1f ~n", [100 * 1 / 30]),
io:fwrite("==================~n").
``` 
```sh
printing with different precisions
====== 5.2 can fail =====
*****
*****
60.00
 3.33
===== 6.2 ok =====
428.57
100.00
 60.00
  3.33
===== 5.1 ok =====
428.6
100.0
 60.0
  3.3
==================
```

I'll try to propose the fix for the `:proper`, and if that won't work in a reasonable time, we can implement a similar printing function [this one and some helper functions below](https://github.com/proper-testing/proper/blob/master/src/proper.erl#L2141)

flowerett:
Hello @alfert, the PR is ready now, everything works for OTP/23
The new functions that appear in the PR are from [this change in PropEr](https://github.com/proper-testing/proper/pull/220/files)
Also, some functions are using PropEr counterparts as [outer_test type is now opaque](https://github.com/proper-testing/proper/commit/df94f4f3e4ad34b283e753d0e5c4cb6da990c659)

alfert:
Hi @flowerett, thanks for that. I will take a look at it soon!

mix.exs
```diff
@@ -78,7 +78,7 @@ defmodule PropCheck.Mixfile do
   # Type `mix help deps` for more examples and options
   defp deps do
     [
-      {:proper, "~> 1.3"},
+      {:proper, "~> 1.3", github: "proper-testing/proper"},
```
alfert, line 81: Shouldn't we fix this to a tested version or do we believe that they will no nasty changes in proper that will break our system?

alfert:
Looks good 💚 ! I will merge, but we need to take a careful look on :proper's future changes breaking our contracts.

mix.exs
flowerett: The last released version is 2y old, and seems there's no other "stable" branch except master.
So we need to be careful and not to update the version locked in the `mix.lock` without a reason.
I guess it's worth a comment.

PR #182, opened 2020-11-04
Misc doc changes

kianmeng:
List of changes:

- Mix format module config
- Use common source url
- Show source reference by tagged version
- Use shorter module name
- Fix incorrect changelog link
- Badges and more badges!

PR #183, opened 2020-11-04
Remove non-running CI jobs

evnu:
This PR is for debugging the zombie CI jobs:

image


.github/workflows/elixir.yml
```diff
@@ -1,84 +1,15 @@
 name: Elixir CI
 
-on:
```
flowerett, line 3: I suppose there's nothing wrong with current jobs defined in .github/workflows/elixir.yml
The expected steps maybe can be cleaned by repo admin.
https://github.community/t/is-there-a-way-to-delete-or-hide-old-renamed-workflows/16281/20
I also saw a long thread about this being an issue, not sure if it is fixed now.

evnu: Yes, this was my impression as well. I found [this](https://github.community/t/github-actions-pull-requests-always-include-some-outdated-checks/16157) as well, which looks similar. @alfert please check the settings and feel free to close this PR if the settings fix this issue.

alfert: There were indeed several checks for master activated which do not fit the current naming scheme in the `elixir.xml` file. I am pretty sure that this is the reason of this annoying task and that we don't have to delete our checks.

alfert: And yes, after deleting the checks, #182 is green - as it should since only documentation was changed.

alfert:
I close this pull request since the solution is to ensure that the required checks on the master branch defined in the repository settings match the names of the actions in the `elixir.yml` file.

PR #185, opened 2020-11-06
Add formatter with default settings

flowerett:
This can be opinionated, but I found that it's way easier to understand and maintain formatted code.
So please consider adding this to the library.

- there are no changes in the codebase, only blank line changes made by `mix format` with default settings.
- these [two lines were also added to the lint command](https://github.com/alfert/propcheck/pull/185/files?diff=unified&w=1#diff-dfa6f4ed74c90e5d4fda283d547d366586e690387289bcfd473e3fa5f9ace308R74):

```
"format --check-formatted",
"deps.unlock --check-unused",
```
so GitHub pipeline should fail if code is not formatted.


lib/basic_types.ex
```diff
@@ -375,7 +374,7 @@ defmodule PropCheck.BasicTypes do
   """
   @spec ordered_list(raw_type()) :: type
   def ordered_list(elem_type) do
-    let l <- list(elem_type), do: :lists.sort(l)
+    let(l <- list(elem_type), do: :lists.sort(l))
```
alfert, line 377: This looks really ugly, does'n it? It appears to me as if for macros always parantheses would be used. Such as 

    for(x <- xs, do: x+1) 

instead of 

    for x <- xs, do: x+1

Can this be handled nicer with the formatter? I never worked with, so I have no experience.

lib/counterstrike.ex
```diff
@@ -80,8 +83,14 @@ defmodule PropCheck.CounterStrike do
   # storing new counter examples.
   @spec load_existing_counter_examples(%{mfa => any}, :dets.tab_name()) :: %{mfa => any}
   defp load_existing_counter_examples(ce, dets) do
-    new_ce = :dets.foldl(fn {mfa, example}, ces ->
-      Map.put_new(ces, mfa, example) end, ce, dets)
+    new_ce =
```
alfert, line 86: while the original source is also not formatted too nicely, but this is quite lengthy, but consistent.

lib/basic_types.ex
flowerett: Yep, I agree, and it can be configured.
So if we're ok with the idea of bringing the formatter in general, I can try to add some rules.

evnu:
@alfert shall we revisit this? Using the formatter would make the code more consistent. I think with `locals_without_parens`, we could already handle most of the more disturbing cases.

lib/counterstrike.ex
evnu: Yes, the formatter sometimes produces ugly code. Sometimes, that actually hints to refactor the code a little bit, which can result in better readability. For this example, maybe something like this would be nice:

```elixir
  defp load_existing_counter_examples(ce, dets) do
    new_ce = :dets.foldl(&add_counterexample/1, ce, dets)

    :ok = :dets.delete_all_objects(dets)
    :ok = :dets.sync(dets)
    new_ce
  end

  defp add_counterexample({mfa, example}, ces) do
    Map.put_new(ces, mfa, example)
  end
```

alfert:
Yeah, I guess we should do it. I tried the formatter in others small projects and I liked what it did to my code (most of the time). This will be a big PR, though, and it is important that the PropCheck specific code becomes not too ugly. So, establishing proper rules (hah!) is important.

evnu:
@flowerett Would you be interested in picking this up again?

evnu:
Closing in favor of #204. Thanks @flowerett!

PR #204, opened 2021-06-03
Add Formatter

evnu:
This picks up the PR by @flowerett (#185). I added `locals_without_parens` for a few common macros (`defcommand`, `let`, `forall`, ..) and prettified one location which I found to be quite ugly after formatting. The PR also includes a check for CI.


flowerett:
Thanks, @evnu!
I guess my PR can be closed then.

alfert (approved):
I like what I see. Not every line looks like I would write it by hand, but alone the overall aligned structure (2 vs. 4. indentations) is very helpful.

alfert:
It helps that I used the formatter in some newer stuff (greenfield), to experience that it is quite useful :-) 

Go for it! 💚

PR #206, opened 2021-08-11
Resolve counter examples in one source code location

### LINKED ISSUE #205: mix propcheck.clean fails on a new project
After creating a new project, `mix propcheck.clean` fails:

```
** (FunctionClauseError) no function clause matching in IO.chardata_to_string/1

    The following arguments were given to IO.chardata_to_string/1:

        # 1
        nil

    Attempted function clauses (showing 2 out of 2):

        def chardata_to_string(string) when is_binary(string)
        def chardata_to_string(list) when is_list(list)

    (elixir 1.12.2) lib/io.ex:615: IO.chardata_to_string/1
    (elixir 1.12.2) lib/file.ex:1088: File.rm/1
    (mix 1.12.2) lib/mix/task.ex:394: anonymous fn/3 in Mix.Task.run_task/3
    (mix 1.12.2) lib/mix/cli.ex:84: Mix.CLI.run_task/2
```

```elixir
efmodule PropcheckCounterexamples.MixProject do
  use Mix.Project

  def project do
    [
      app: :propcheck_counterexamples,
      version: "0.1.0",
      elixir: "~> 1.12",
      start_permanent: Mix.env() == :prod,
      deps: deps()
    ]
  end

  # Run "mix help compile.app" to learn about applications.
  def application do
    [
      extra_applications: [:logger]
    ]
  end

  # Run "mix help deps" to learn about dependencies.
  defp deps do
    [
      {:propcheck, "~> 1.4"}
    ]
  end
end
```

@alfert I think [this](https://github.com/alfert/propcheck/blob/master/lib/mix.ex#L52) needs to take the default counter examples path into account.

evnu:
This pull request refactors resolving the location of the counter examples file. Previously, `mix propcheck.clean` differed slightly in how it resolved that file compared to `app.ex`. With this pull request, only one location determines where to look for counter examples (`Mix.PropCheck`).

Fix #205

### DIFF
--- a/lib/app.ex
+++ b/lib/app.ex
@@ -49,7 +49,7 @@ defmodule PropCheck.App do
 
   defp counter_example_file do
     case Application.get_env(:propcheck, :counter_examples) do
-      nil -> Mix.counter_example_file() || Mix.default_counter_examples_file()
+      nil -> Mix.counter_example_file_or_default()
       counter_example_file -> counter_example_file
     end
   end
--- a/lib/app.ex
+++ b/lib/app.ex
@@ -32,7 +32,7 @@ defmodule PropCheck.App do
   defp populate_application_env do
     Application.put_env(:propcheck, :global_verbose, global_verbose())
     Application.put_env(:propcheck, :global_detect_exceptions, global_detect_exceptions())
-    Application.put_env(:propcheck, :counter_example_file, counter_example_file())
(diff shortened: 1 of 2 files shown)


evnu:
@alfert what do you think?

alfert:
Looks good and was a bad idea in the first place.

PR #209, opened 2021-11-02
Do not interpret [] as a command list

### LINKED ISSUE #208: Counter-Example is empty when test failed
Counter-Example output is empty.
I am expecting it return since there is a bug in the function biggest

mix test
..

  1) property biggest num test (PbtTest)
     test/pbt_test.exs:10
     Property Elixir.PbtTest.property biggest num test() failed. Counter-Example is:

     Counter example stored.

     code: nil
     stacktrace:
       (propcheck 1.4.0) lib/properties.ex:227: PropCheck.Properties.handle_check_results/2
       test/pbt_test.exs:10: (test)

Finished in 0.1 seconds (0.00s async, 0.1s sync)
1 doctest, 1 property, 1 test, 1 failure

`
defmodule PbtTest do
  use ExUnit.Case
  doctest Pbt
  use PropCheck

  test "greets the world" do
    assert Pbt.hello() == :world
  end

  property "biggest num test" do
    forall(type <- :proper_types.list(:proper_types.integer())) do
      biggest(type) == List.last(Enum.sort(type))
    end
  end

  def biggest([head | tail]) do
    biggest(tail, head)
  end

  defp biggest([], max) do
    max
  end

  defp biggest([head | tail], max) when head > max do
    biggest(tail, head)
  end

  defp biggest([head | tail], max) when head < max do
    biggest(tail, max)
  end
end

`
elixir -v
Erlang/OTP 24 [erts-12.0.2] [source] [64-bit] [smp:16:16] [ds:16:16:10] [async-threads:1] [jit] [dtrace]

Elixir 1.12.1 (compiled with Erlang/OTP 24)

 {:propcheck, "~> 1.4", only: [:test, :dev]}

evnu:
`is_statem_commands/1` interprets `[]` as a command list, but a command list should have at least a `:init` tuple. Due to this misinterpretation, an incorrect pretty printer is used, which results in a counter example of `[]` not being printed.

Fix #208

### DIFF
--- a/lib/properties.ex
+++ b/lib/properties.ex
@@ -295,6 +295,10 @@ defmodule PropCheck.Properties do
     end
   end
 
+  defp is_statem_commands([[]]) do
+    :data
+  end
+
   defp is_statem_commands([counter_example]) when is_list(counter_example) do
     is_command_list =
       counter_example


alfert:
What a nice glitch on quantification of empty enums. From the `Enum.all?` examples: 

```
iex> Enum.all?([], fn _ -> nil end)
true
```

Thanks for resolving that.

alfert (approved):

alfert:
@evnu should we go for a 1.4.1 release?

evnu:
> @evnu should we go for a 1.4.1 release?

Yes, I think that makes sense.

alfert:
Release created.

PR #212, opened 2022-01-27
Fix module name in deprecation warning

aaronrenner:
I was reading the docs and found that clicking the link in the deprecation warning for `PropCheck.StateM.DSL` didn't work. I believe this is because of a capitalization issue in the module name.


alfert:
Thanks a lot! I use Dash on Mac, there the capitalisation was not an issue. But it will on every Unix filesystem.

PR #218, opened 2023-04-28
Fixes documentation of `run_commands`

adkron:
The `PropCheck.StateM.ModelDSL` documentation shows `run_commands(cmds)` to run the commands, but there is no single argument version of `run_commands/2`. The first argument is the module that implements the callbacks.

I think that this is left over from `PropCheck.StateM.DSL`


alfert:
Thank you Amos!

PR #220, opened 2023-05-02
Fix gh action

alfert:


evnu:
See https://github.com/alfert/propcheck/pull/224 for another test to fix the CI.

evnu:
#224 fixes PropCheck for newer versions of Elixir and OTP.

PR #223, opened 2023-10-05
ci: Update versions and fix action

evnu:
This PR removes obsolete Elixir versions and fixes the CI pipeline.


evnu:
Closing, I only now saw that @alfert already started with https://github.com/alfert/propcheck/pull/220. I think we are somewhat blocked on a missing new PropEr version. @alfert maybe we should reference a commit for PropEr, they don't seem to release versions at all anymore.

PR #225, opened 2024-03-15
chore: removed redundant application declaration

iToXiQ:
There is no need to declare proper and libgraph in applications because they will be derived via deps automatically.
Currently you will also get the following warning when compiling propcheck.

`both :extra_applications and :applications was found in your mix.exs. You most likely want to remove the :applications key, as all applications are derived from your dependencies`

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -59,9 +59,8 @@ defmodule PropCheck.Mixfile do
   # Type `mix help compile.app` for more information
   def application do
     [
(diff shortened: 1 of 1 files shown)


evnu:
Thanks! @alfert don't know if this warrants a Patch release to make the warning go away for users.

alfert:
Let's do a patch release, feels like a nice service for users.

PR #224, opened 2023-10-20
Fix Github Actions

evnu:
Another run at fixing CI. This replaces the github action versions and restricts the checks to recent versions of OTP and Elixir. The PR also bumps dependencies to the current versions.

PR #226, opened 2024-03-16
Set proper to the official 1.4 release

alfert:


alfert:
@evnu Any ideas why this fails? It does so also locally on my machine. But I was too long away from Elixir to have a good intuition here.

evnu:
@alfert I think I saw that error with proper some time ago. Currently on mobile, but this might give a hint: https://github.com/proper-testing/proper/blob/a5ae5669f01143b0828fc21667d4f5e344aa760b/src/proper_gen.erl#L361 was there a version bump in proper that we need?

evnu:
@alfert I thought we fixed the eunit error with this: https://github.com/alfert/propcheck/pull/224

evnu:
@alfert ok, so the current proper release on hex.pm is just too old. That's why we used a git revision.

alfert:
This is bad, since we cannot create a proper hex release with dependency on a git hash :-(

evnu:
@alfert so time to ask for a PropEr release then I guess 🙈

evnu:
Seems that there is no new release coming for PropEr any time soon. :-/

evnu:
Forking would mean to release PropEr to `hex.pm`, I think that's probably more work than we want to have for `PropCheck` :) Can we vendor-in PropEr and have that as part of this source tree?

alfert:
As a git submodule?

PR #227, opened 2024-06-13
fix: use ExUnit.Case.register_test/6 rather than ExUnit.Case.register_test/4

ahamez:
ExUnit.Case.register_test/4 is deprecated since Elixir 1.12

### DIFF
--- a/lib/properties.ex
+++ b/lib/properties.ex
@@ -114,7 +114,16 @@ defmodule PropCheck.Properties do
 
       # @tag failing_prop: tag_property({module, prop_name, []})
       tags = [[failing_prop: tag_property({module, name, []})]]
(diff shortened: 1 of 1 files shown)


evnu:
Thanks. Can you format the file? :)

ahamez:
Done!

ahamez:
Hello 👋 Would it possible to release a new version with this fix? 🙏

michal-stx:
I second the question :) Could you please release a new hex package with this fix?

evnu:
> Could you please release a new hex package with this fix?

Alas, we are unable to cut a new release before [PropEr](https://github.com/proper-testing/proper) creates a new one first. See https://github.com/proper-testing/proper/issues/305 and https://github.com/alfert/propcheck/pull/226.

